PR #2, opened 2023-12-29
Sort weather stations

nipafx:
I sorted the weather stations but LibreOffice Calc didn't quite hit the Java String sort order, so some small fixes remain (e.g. "Abéché" shouldn't be 1st but 3rd) - I'll leave those to others. 😁


gunnarmorling (approved):

gunnarmorling:
Excellent, thanks!

PR #4, opened 2024-01-01
Initial implementation, just to get something in, parallel processing using records

royvanrijn:
First of all: Nice work running such a competition, love it.

My implementation is: [CalculateAverage_royvanrijn.java]

Using the default installed JDK on my machine (21.0.1-zulu) on an M2 Macbook Pro (32 GB), it runs in +/- 0m54.437s

### DIFF
--- a/calculate_average_royvanrijn.sh
+++ b/calculate_average_royvanrijn.sh
@@ -17,4 +17,4 @@
 
 
 JAVA_OPTS=""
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Very nice, thank you so much for submitting something so quickly! Can confirm it's definitely faster than the baseline, congrats for taking over the leaderboard :) Running it as we speak, will update the board in a bit.

gunnarmorling:
Squashed the two commits into one and merged. Thanks again, @royvanrijn! Congrats on taking over the [leaderboard](https://github.com/gunnarmorling/1brc/blob/main/README.md#results) :)

PR #5, opened 2024-01-01
memory mapped files, branchless parsing, bitwiddle magic

royvanrijn:
I've ran it using:
`sdk use java 21.0.1-graal`

Latest graalvm seems to give the best performance.

I've added AppCDS, very simple, by dumping on the first run, using on the other runs.

I've also implemented memory mapped files, based on spullara/bjhara's submission, that's awesome and gave even more improvement ideas.

Most improvements come from optimizing the inner-most loop (of course) by writing my own branchless parser for the numbers, which probably looks like magic.

Finally I've added a custom HashTable implementation, just for this specific usecase.... speeeed.

On the MacBook Pro M2 (32GB) it runs in:
`real	0m2.873s`

### DIFF
--- a/calculate_average_royvanrijn.sh
+++ b/calculate_average_royvanrijn.sh
@@ -16,5 +16,11 @@
 #
 
 
-JAVA_OPTS=""
+# Added for fun, doesn't seem to be making a difference...
+if [ -f "target/calculate_average_royvanrijn.jsa" ]; then
+    JAVA_OPTS="-XX:SharedArchiveFile=target/calculate_average_royvanrijn.jsa -Xshare:on"
+else
+    # First run, create the archive:
+    JAVA_OPTS="-XX:ArchiveClassesAtExit=target/calculate_average_royvanrijn.jsa"
+fi
 time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_royvanrijn
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
@@ -15,309 +15,65 @@
  */
 package dev.morling.onebrc;
 
-import java.io.File;
 import java.io.IOException;
-import java.io.RandomAccessFile;
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.StandardOpenOption;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.TreeMap;
+import java.util.AbstractMap;
+import java.util.Map;
 import java.util.stream.Collectors;
 
-/**
- * Changelog:
- *
- * Initial submission:          62000 ms
- * Chunked reader:              16000 ms
- * Optimized parser:            13000 ms
- * Branchless methods:          11000 ms
- * Adding memory mapped files:  6500 ms (based on bjhara's submission)
- * Skipping string creation:    4700 ms
- * Custom hashmap...            4200 ms
- * Added SWAR token checks:     3900 ms
- * Skipped String creation:     3500 ms (idea from kgonia)
- * Improved String skip:        3250 ms
- * Segmenting files:            3150 ms (based on spullara's code)
- * Not using SWAR for EOL:      2850 ms
- *
- * Best performing JVM on MacBook M2 Pro: 21.0.1-graal
- * `sdk use java 21.0.1-graal`
- *
- */
 public class CalculateAverage_royvanrijn {
 
     private static final String FILE = "./measurements.txt";
 
-    // mutable state now instead of records, ugh, less instantiation.
-    static final class Measurement {
-        int min, max, count;
-        long sum;
+    private record Measurement(double min, double max, double sum, long count) {
 
-        public Measurement() {
-            this.min = 10000;
-            this.max = -10000;
+        Measurement(double initialMeasurement) {
+            this(initialMeasurement, initialMeasurement, initialMeasurement, 1);
         }
 
-        public Measurement updateWith(int measurement) {
-            min = min(min, measurement);
-            max = max(max, measurement);
-            sum += measurement;
-            count++;
-            return this;
-        }
-
-        public Measurement updateWith(Measurement measurement) {
-            min = min(min, measurement.min);
-            max = max(max, measurement.max);
-            sum += measurement.sum;
-            count += measurement.count;
-            return this;
+        public static Measurement combineWith(Measurement m1, Measurement m2) {
+            return new Measurement(
+                    m1.min < m2.min ? m1.min : m2.min,
+                    m1.max > m2.max ? m1.max : m2.max,
+                    m1.sum + m2.sum,
+                    m1.count + m2.count
+            );
         }
 
         public String toString() {
-            return round(min) + "/" + round((1.0 * sum) / count) + "/" + round(max);
+            return round(min) + "/" + round(sum / count) + "/" + round(max);
         }
 
         private double round(double value) {
-            return Math.round(value) / 10.0;
+            return Math.round(value * 10.0) / 10.0;
         }
     }
 
-    public static final void main(String[] args) throws Exception {
-        new CalculateAverage_royvanrijn().run();
+    public static void main(String[] args) throws IOException {
+
+        // long before = System.currentTimeMillis();
+
+        Map<String, Measurement> resultMap = Files.lines(Path.of(FILE)).parallel()
+                .map(record -> {
+                    // Map to <String,double>
+                    int pivot = record.indexOf(";");
+                    String key = record.substring(0, pivot);
+                    double measured = Double.parseDouble(record.substring(pivot + 1));
+                    return new AbstractMap.SimpleEntry<>(key, measured);
+                })
+                .collect(Collectors.toConcurrentMap(
+                        // Combine/reduce:
+                        AbstractMap.SimpleEntry::getKey,
+                        entry -> new Measurement(entry.getValue()),
+                        Measurement::combineWith));
+
+        System.out.print("{");
+        System.out.print(
+                resultMap.entrySet().stream().sorted(Map.Entry.comparingByKey()).map(Object::toString).collect(Collectors.joining(", ")));
+        System.out.println("}");
+
+        // System.out.println("Took: " + (System.currentTimeMillis() - before));
+
     }
-
-    private void run() throws Exception {
-
-        var results = getFileSegments(new File(FILE)).stream().map(segment -> {
-
-                    long segmentEnd = segment.end();
-                    try (var fileChannel = (FileChannel) Files.newByteChannel(Path.of(FILE), StandardOpenOption.READ)) {
-                        var bb = fileChannel.map(FileChannel.MapMode.READ_ONLY, segment.start(), segmentEnd - segment.start());
-                        var buffer = new byte[64];
-
-                        // Force little endian:
-                        bb.order(ByteOrder.LITTLE_ENDIAN);
-
-                        BitTwiddledMap measurements = new BitTwiddledMap();
-
-                        int startPointer;
-                        int limit = bb.limit();
-                        while ((startPointer = bb.position()) < limit) {
-
-                            // SWAR is faster for ';'
-                            int separatorPointer = findNextSWAR(bb, SEPARATOR_PATTERN, startPointer + 3, limit);
-
-                            // Simple is faster for '\n' (just three options)
-                            int endPointer;
-                            if (bb.get(separatorPointer + 4) == '\n') {
-                                endPointer = separatorPointer + 4;
-                            }
-                            else if (bb.get(separatorPointer + 5) == '\n') {
-                                endPointer = separatorPointer + 5;
-                            }
-                            else {
-                                endPointer = separatorPointer + 6;
-                            }
-
-                            // Read the entry in a single get():
-                            bb.get(buffer, 0, endPointer - startPointer);
-                            bb.position(endPointer + 1); // skip to next line.
-
-                            // Extract the measurement value (10x):
-                            final int nameLength = separatorPointer - startPointer;
-                            final int valueLength = endPointer - separatorPointer - 1;
-                            final int measured = branchlessParseInt(buffer, nameLength + 1, valueLength);
-                            measurements.getOrCreate(buffer, nameLength).updateWith(measured);
-                        }
-                        return measurements;
-                    }
-                    catch (IOException e) {
-                        throw new RuntimeException(e);
-                    }
-                }).parallel().flatMap(v -> v.values.stream())
-                .collect(Collectors.toMap(e -> new String(e.key), BitTwiddledMap.Entry::measurement, (m1, m2) -> m1.updateWith(m2), TreeMap::new));
-
-        // Seems to perform better than actually using a TreeMap:
-        System.out.println(results);
-    }
-
-    /**
-     * -------- This section contains SWAR code (SIMD Within A Register) which processes a bytebuffer as longs to find values:
-     */
-    private static final long SEPARATOR_PATTERN = compilePattern((byte) ';');
-
-    private int findNextSWAR(ByteBuffer bb, long pattern, int start, int limit) {
-        int i;
-        for (i = start; i <= limit - 8; i += 8) {
-            long word = bb.getLong(i);
-            int index = firstAnyPattern(word, pattern);
-            if (index < Long.BYTES) {
-                return i + index;
-            }
-        }
-        // Handle remaining bytes
-        for (; i < limit; i++) {
-            if (bb.get(i) == (byte) pattern) {
-                return i;
-            }
-        }
-        return limit; // delimiter not found
-    }
-
-    private static long compilePattern(byte value) {
-        return ((long) value << 56) | ((long) value << 48) | ((long) value << 40) | ((long) value << 32) |
-                ((long) value << 24) | ((long) value << 16) | ((long) value << 8) | (long) value;
-    }
-
-    private static int firstAnyPattern(long word, long pattern) {
-        final long match = word ^ pattern;
-        long mask = match - 0x0101010101010101L;
-        mask &= ~match;
-        mask &= 0x8080808080808080L;
-        return Long.numberOfTrailingZeros(mask) >>> 3;
-    }
-
-    record FileSegment(long start, long end) {
-    }
-
-    /** Using this way to segment the file is much prettier, from spullara */
-    private static List<FileSegment> getFileSegments(File file) throws IOException {
-        final int numberOfSegments = Runtime.getRuntime().availableProcessors();
-        final long fileSize = file.length();
-        final long segmentSize = fileSize / numberOfSegments;
-        final List<FileSegment> segments = new ArrayList<>();
-        try (RandomAccessFile randomAccessFile = new RandomAccessFile(file, "r")) {
-            for (int i = 0; i < numberOfSegments; i++) {
-                long segStart = i * segmentSize;
-                long segEnd = (i == numberOfSegments - 1) ? fileSize : segStart + segmentSize;
-                segStart = findSegment(i, 0, randomAccessFile, segStart, segEnd);
-                segEnd = findSegment(i, numberOfSegments - 1, randomAccessFile, segEnd, fileSize);
-
-                segments.add(new FileSegment(segStart, segEnd));
-            }
-        }
-        return segments;
-    }
-
-    private static long findSegment(int i, int skipSegment, RandomAccessFile raf, long location, long fileSize) throws IOException {
-        if (i != skipSegment) {
-            raf.seek(location);
-            while (location < fileSize) {
-                location++;
-                if (raf.read() == '\n')
-                    return location;
-            }
-        }
-        return location;
-    }
-
-    /**
-     * Branchless parser, goes from String to int (10x):
-     * "-1.2" to -12
-     * "40.1" to 401
-     * etc.
-     *
-     * @param input
-     * @return int value x10
-     */
-    private static int branchlessParseInt(final byte[] input, int start, int length) {
-        // 0 if positive, 1 if negative
-        final int negative = ~(input[start] >> 4) & 1;
-        // 0 if nr length is 3, 1 if length is 4
-        final int has4 = ((length - negative) >> 2) & 1;
-
-        final int digit1 = input[start + negative] - '0';
-        final int digit2 = input[start + negative + has4];
-        final int digit3 = input[start + negative + has4 + 2];
-
-        return (-negative ^ (has4 * (digit1 * 100) + digit2 * 10 + digit3 - 528) - negative); // 528 == ('0' * 10 + '0')
-    }
-
-    // branchless max (unprecise for large numbers, but good enough)
-    static int max(final int a, final int b) {
-        final int diff = a - b;
-        final int dsgn = diff >> 31;
-        return a - (diff & dsgn);
-    }
-
-    // branchless min (unprecise for large numbers, but good enough)
-    static int min(final int a, final int b) {
-        final int diff = a - b;
-        final int dsgn = diff >> 31;
-        return b + (diff & dsgn);
-    }
-
-    /**
-     * A normal Java HashMap does all these safety things like boundary checks... we don't need that, we need speeeed.
-     *
-     * So I've written an extremely simple linear probing hashmap that should work well enough.
-     */
-    class BitTwiddledMap {
-        private static final int SIZE = 16384; // A bit larger than the number of keys, needs power of two
-        private int[] indices = new int[SIZE]; // Hashtable is just an int[]
-
-        BitTwiddledMap() {
-            // Optimized fill with -1, fastest method:
-            int len = indices.length;
-            if (len > 0) {
-                indices[0] = -1;
-            }
-            // Value of i will be [1, 2, 4, 8, 16, 32, ..., len]
-            for (int i = 1; i < len; i += i) {
-                System.arraycopy(indices, 0, indices, i, i);
-            }
-        }
-
-        private List<Entry> values = new ArrayList<>(512);
-
-        record Entry(int hash, byte[] key, Measurement measurement) {
-            @Override
-            public String toString() {
-                return new String(key) + "=" + measurement;
-            }
-        }
-
-        /**
-         * Who needs methods like add(), merge(), compute() etc, we need one, getOrCreate.
-         * @param key
-         * @return
-         */
-        public Measurement getOrCreate(byte[] key, int length) {
-            int inHash;
-            int index = (SIZE - 1) & (inHash = hashCode(key, length));
-            int valueIndex;
-            Entry retrievedEntry = null;
-            while ((valueIndex = indices[index]) != -1 && (retrievedEntry = values.get(valueIndex)).hash != inHash) {
-                index = (index + 1) % SIZE;
-            }
-            if (valueIndex >= 0) {
-                return retrievedEntry.measurement;
-            }
-            // New entry, insert into table and return.
-            indices[index] = values.size();
-
-            // Only parse this once:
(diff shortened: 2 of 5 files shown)


manakbisht:
How about disabling the gc?
`JAVA_OPTS="-Xms16g -Xmx24g -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC"`
For a large enough input size, the initial heap allocation time should be lesser than the time taken for garbage collection. Unfortunately, I don't have the specs to test this.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -50,30 +63,99 @@ private double round(double value) {
...
+        }
+
+        // Combine results from all futures, is there a better way to do this?
+        Map<String, Measurement> combinedResult = new TreeMap<>();
+        for (Future<Map<String, Measurement>> future : futures) {
+            final Map<String, Measurement> resultMap = future.get();
+            for (Map.Entry<String, Measurement> entry : resultMap.entrySet()) {
+                combinedResult.merge(entry.getKey(), entry.getValue(), Measurement::combineWith);
+            }
+        }
```
qtxo, line 97: > // Combine results from all futures, is there a better way to do this?

Maybe a bit more elegant using ExecutorCompletionService

```
var cs = new ExecutorCompletionService<Map<String, Measurement>>(executor);

long lastEnd = 0;
for (int i = 0; i < numberOfChunks; i++) {
	// Method findNextLineStart finds the correct point to break;
	long start = lastEnd;
	long end = (i == numberOfChunks - 1) ? fileSize : findNextLineStart(FILE, (i + 1) * chunkSize);
	cs.submit(() -> processChunk(FILE, start, end));
	lastEnd = end;
}

for (int i = 0; i < numberOfChunks; i++) {
	for (var entry : cs.take().get().entrySet()) {
		combinedResult.merge(entry.getKey(), entry.getValue(), Measurement::combineWith);
	}
}
```

royvanrijn: Ah yes, that is better, doesn't improve performance, but feels better :)

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -50,30 +63,99 @@ private double round(double value) {
...
     }
 
-    public static void main(String[] args) throws IOException {
+    public static void main(String[] args) throws Exception {
+
+        long before = System.currentTimeMillis();
+
+        // Split the file into numberOfChunks, use random access to read and process in parallel.
+
+        int numberOfChunks = 42; // Of course.
```
keshavram-apptware, line 72: Why is 42 obvious?

royvanrijn: It is the answer to life, the universe and everything.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,182 @@
...
+            }
+            return measurements;
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
 
+    // Specialized base-10 parser, skips - and .
+    private static int branchlessParseInt(final String input) {
```
royvanrijn, line 156: Don't ask...

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,182 @@
...
+        // "-1.2" to -12
+        // "40.1" to 401
+        // etc.
+
+        // Magic.
+        int off = ~(input.charAt(0) >> 4) & 1;
+        int neg = 1 - off - off;
+        int has4 = ((input.length() - off) >> 2) & 1;
+        return neg * ((input.charAt(off) - '0') * 100 * has4 + (input.charAt(off + has4) - '0') * 10 + (input.charAt(2 + off + has4) - '0'));
+    }
```
royvanrijn, line 167: Don't ask me about this code, its magic.

It takes these numbers in the following formats:
-X.X
-XX.X
X.X
XX.X

and turns them into integers XXX (10x the original amount) without any branches.

Turns out chatAt() is faster in this case than getting the array and referencing that.

royvanrijn: I think I've tried that, it is something the compiler instantly picks up anyway, it knows these variables are final and I didn't see a performance difference. They probably get smooshed into a couple/single instructions anyway since they have such a limited scope.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,182 @@
...
+
+    public static final void main(String[] args) throws Exception {
+
+        // Split the file into numberOfChunks, use random access to read and process in parallel.
+
+        final long fileSize = new File(FILE).length();
+        final long chunkSize = fileSize / CHUNKS;
+
+        final var es = Executors.newVirtualThreadPerTaskExecutor();
```
qtxo, line 94: I think VT are not well suited for file IO. Have you tried a regular executor? In my case it is faster.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -16,64 +16,224 @@
...
+    /**
+     * Branchless parser, goes from String to int (10x):
+     * "-1.2" to -12
+     * "40.1" to 401
+     * etc.
+     *
+     * @param input
+     * @return int value x10
+     */
+    private static int branchlessParseInt(final byte[] input, int length) {
```
DamienOReilly, line 213: I wonder if you would get any performance improvements caching the temperature to int? Assuming temperature ranges from -99.9 to 99.9, there would only be at most ~2000 keys?

swaechter:
@royvanrijn Nice solution, I had something similar in mind. I think if using FFI for mmap and pthreads would improve the performance (Less byte arrays for buffering etc.). Maybe I'll find the time :)

suchwerk:
What about using integer arithmetic instead of floats?

gunnarmorling:
Hey @royvanrijn, are you planning to do further changes to this one? If so, wanna put it into "Draft" state until it's good to go from your side?

gunnarmorling:
@royvanrijn, could you rebase this one to current main and squash everything into a single commit?

calculate_average_royvanrijn.sh
```diff
@@ -16,5 +16,11 @@
...
-JAVA_OPTS=""
-time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_royvanrijn
+# Added for fun, doesn't seem to be making a difference...
+if [ -f "target/calculate_average_royvanrijn.jsa" ]; then
+    JAVA_OPTS="-XX:SharedArchiveFile=target/calculate_average_royvanrijn.jsa -Xshare:on"
+else
+    # First run, create the archive:
+    JAVA_OPTS="-XX:ArchiveClassesAtExit=target/calculate_average_royvanrijn.jsa"
+fi
+java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_royvanrijn
```
gunnarmorling, line 26: Could you bring back the time call here? I figured it saves a few ms to have it here vs. in the evaluate loop script, so let's do it here in the actual scripts for the sake of consistency and comparability.

gunnarmorling:
Preliminary updated the leaderboard with the result from the latest version here: 00:23.366 on the evaluation environment. You've taken over again, @royvanrijn :)

calculate_average_royvanrijn.sh
royvanrijn: I've re-added the `time` there and squashed everything after a long and hard battle.

Also improved the code a little with a tiny custom HashTable, because, why not? Speeeed.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -16,64 +16,319 @@
...
+            public boolean hasNext() {
+                return bytesRead < size;
+            }
+        }, Spliterator.IMMUTABLE), false);
+    }
+
+    /**
+     * Branchless parser, goes from String to int (10x):
+     * "-1.2" to -12
+     * "40.1" to 401
```
AlexanderYastrebov, line 246: @gunnarmorling Is this assumption acceptable? E.g. `1`, `0`, `2.00` are all valid doubles.

AlexanderYastrebov: Also there should be an acceptance test suite for the implementations, I am pretty sure this implementation does not produce the same output as the baseline.

royvanrijn: I'm pretty sure it produces the exact same output, I check regularly with each change.

The input and output have one decimal place precision (as stated by the website "rounded to one fractional digit").

I'm internally storing the doubles as 10x integers because the precision is just a single digit, and I'm making sure the rounding it correct afterwards for the average.

AlexanderYastrebov: The README only says about output format but not input.
https://github.com/gunnarmorling/1brc/blob/e7e7deb64a9d1ace22e5f9091ba7c06d1d7f1cd4/README.md?plain=1#L27-L28
so its worth clarifying.

There are many rounding modes - this is also not specified, e.g. in go https://pkg.go.dev/math#Round is not the same as in java.

AlexanderYastrebov: Also reference implementation does ad-hoc rounding `Math.round(value * 10.0) / 10.0` but actual output depends on the string concatenation which performs another rounding, see https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Double.html#toString(double)

AlexanderYastrebov: > I am pretty sure this implementation does not produce the same output as the baseline.
> I'm pretty sure it produces the exact same output, I check regularly with each change.

To prove the point here an input https://gist.github.com/AlexanderYastrebov/d560d697c61484cc5eedaa09120d6464 for which your implementation produces different output than the baseline.

Due to unfortunate output format (see https://github.com/gunnarmorling/1brc/issues/14) one hs to use word diff:
```diff
$ git --no-pager diff --word-diff /tmp/baseline.txt /tmp/royvanrijn.txt
diff --git a/tmp/baseline.txt b/tmp/royvanrijn.txt
index 1a3d2ea..5c739b7 100644
--- a/tmp/baseline.txt
+++ b/tmp/royvanrijn.txt
@@ -1 +1 @@
<omitted for brevity>, Yakutsk=3.4/3.4/3.4, [-Zagreb=7.7/12.5/17.4}-]{+Zagreb=7.7/12.6/17.4}+}
```

You can find such inputs yourself via
```sh
$ while ./create_measurements.sh 100 && diff <(./calculate_average_royvanrijn.sh) <(./calculate_av
erage.sh) ; do echo OK; done
```

AlexanderYastrebov: > Also reference implementation does ad-hoc rounding Math.round(value * 10.0) / 10.0 but actual output depends on the string concatenation which performs another rounding

The baseline rounding is not always correct as could be shown in java playground https://dev.java/playground/ via
```java
double value = (33.6 + 31.7 + 21.9 + 14.6) / 4.0; // 101.8/4 == 25.45 exactly but can not be represented as double; it is 25.449999999999996
double rounded = Math.round(value * 10.0) / 10.0; // 25.45 * 10 == 254.5, round(254.5) == 255, 255/10 should be 25.5

System.out.println("==" + rounded + "=="); // but prints ==25.4==
```

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -16,64 +16,319 @@
...
     }
+
+    /**
+     * -------- This section contains SWAR code (SIMD Within A Register) which processes a bytebuffer as longs to find values:
+     */
+
+    private int findNextSWAR(ByteBuffer bb, long pattern, int start, int limit) {
+        int i;
+        for (i = start; i <= limit - 8; i += 8) {
+            long word = bb.getLong(i);
```
franz1981, line 175: By default ByteBuffers uses Big Indian, which is native for M1/M2...but not on x86, which is the one used for the benchmark.
It translates in:
- wrong position, while found
- a useless reverse bytes 

It would be better to store in a static final, the native byte order and based on it, correctly reverse the bytes yourself (look at what I have done in the Netty code).
Conversely you can always assume little endian, setting the mapped byte buffer order, and leave M1 to have the performance hit

royvanrijn: Ugh...

Wouldn't it be sufficient to change the numberLeadingZero's to numberTrailingZero's? I believe that's what I Netty's code did. But I'm probably failing to see the difference between getLong/getLongLE?

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
royvanrijn: Ah, you are correct in that example, my code would correctly return 25.5 in that last one instead of 25.4.
That's a bit of a pain...

Let's see what @gunnarmorling says about the rounding.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
franz1981: Yep, It Is, but the choice about which order to set the buffer in order to achieve better performance (using native order and save reverse performed by ByteBuffer) force you to add that case to handle both. If you don't care about having a Long.reverse (which is still a single ASM instruction in x86) you can ignore it.
In case you have to add that condition, I suggest to make it static final, which will avoid having the actual condition to happen at runtime.

royvanrijn: I've decided to just go with LE for now and change the ByteBuffer, that way it should just work.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,309 @@
...
+ * Chunked reader:              16000 ms
+ * Optimized parser:            13000 ms
+ * Branchless methods:          11000 ms
+ * Adding memory mapped files:  6500 ms (based on bjhara's submission)
+ * Skipping string creation:    4700 ms
+ * Custom hashmap...            4200 ms
+ * Added SWAR token checks:     3900 ms
+ * Skipped String creation:     3500 ms (idea from kgonia)
+ * Improved String skip:        3250 ms
+ * Segmenting files:            3150 ms (based on spullara's code)
```
franz1981, line 45: I believe EOL is small enough distant that it is not worthy, unless you first check for it.
A trick to make EOL faster is to explode the loop search , as I have done in Netty (that was a the unrolled version for the tail checks < 8 bytes) - that was the reason for that, and was the reason why the unrolled version was coming "before" the batchy one ie to speedup < 8 bytes length cases

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
gunnarmorling: > The README only says about output format but not input

Ah yes. For the sake of simplicity, it's to assume that each input value also has exactly one fractional digit. I'll clarify that in the README.

> Let's see what @gunnarmorling says about the rounding.

Sorry for not being explicit enough here. Can the behavior of the reference implementation be described using any of the existing values of [RoundingMode](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/math/RoundingMode.html)? E.g. `FLOOR`? If so, I'd suggest we make this explicit and any implementation would have to conform to that.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
franz1981: I see you already have an enrolled check there, so I think my previous comment is unnecessary

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,309 @@
...
+        for (i = start; i <= limit - 8; i += 8) {
+            long word = bb.getLong(i);
+            int index = firstAnyPattern(word, pattern);
+            if (index < Long.BYTES) {
+                return i + index;
+            }
+        }
+        // Handle remaining bytes
+        for (; i < limit; i++) {
+            if (bb.get(i) == (byte) pattern) {
```
franz1981, line 166: Not the pattern here but the actual delimiter, but is a minor really

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,309 @@
...
+            // Only parse this once:
+            byte[] actualKey = new byte[length];
+            System.arraycopy(key, 0, actualKey, 0, length);
+
+            Entry toAdd = new Entry(inHash, actualKey, new Measurement());
+            values.add(toAdd);
+            return toAdd.measurement;
+        }
+
+        private static int hashCode(byte[] a, int length) {
```
franz1981, line 314: Hash code here has a data dependency: you either manually unroll this or just relax the hash code by using a car handle and use getLong amortizing the data dependency in batches, handing only the last 7 (or less) bytes separately, using the array

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
gunnarmorling: But that's the thing, I don't think we can change the behavior of the reference implementation at this point, as it would render existing submissions invalid if they implement a different behavior. So I'd rather make the behavior of the RI explicit, also if it's not the most natural one (agreed that `HALF_UP` behavior would have been better).

gunnarmorling:
@royvanrijn, so what should we do with this one, and all the pending discussions? Wanna submit it as is and keep honing in follow-up PRs? I think it would be nice to be able to update the leaderboard with the current status (fastest right now is @spullara). For that, could you rebase it to resolve the merge conflict?

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
AlexanderYastrebov: I don't think its possible to fix RI because it uses double division and rounds twice.
Since there are no acceptance tests I bet a lot of implementations (those that do not parse and calculate values the same way) will not match RI anyways.

I think RI should favor correctness over performance, then it can be used to build acceptance test suite.

gunnarmorling: Ok, I've logged https://github.com/gunnarmorling/1brc/issues/49 for getting this one sorted out separately and get this PR merged. Let's continue the rounding topic over there. Thx!

gunnarmorling:
@royvanrijn, could you add a line [like this](https://github.com/gunnarmorling/1brc/pull/45) to your launch script for making sure the evaluation is done with GraalVM? I'll squash everything then and evaluate. Thx!

gunnarmorling:
@royvanrijn dang, so I've done what I shouldn't have done and merged it _before_ running. But it seems to take much longer / hang now actually. Any idea what's wrong?

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,309 @@
...
+        return ((long) value << 56) | ((long) value << 48) | ((long) value << 40) | ((long) value << 32) |
+                ((long) value << 24) | ((long) value << 16) | ((long) value << 8) | (long) value;
+    }
+
+    private static int firstAnyPattern(long word, long pattern) {
+        final long match = word ^ pattern;
+        long mask = match - 0x0101010101010101L;
+        mask &= ~match;
+        mask &= 0x8080808080808080L;
+        return Long.numberOfTrailingZeros(mask) >>> 3;
```
franz1981, line 183: Here is shouldn't be the number of leading ones?

May that is why it hasn't @royvanrijn @gunnarmorling

franz1981:
@gunnarmorling added a comment to help

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
royvanrijn: Weird, I thought by setting explicitly to LE would make the compatibility issues disappear. So running it on my machine would automatically mean it works on the target, although perhaps having a performance hit.

afk atm, I’ll check tomorrow, if somebody wants to fix it and tell me, be my guest 😂

franz1981: This method is one of the few cases I suggest to add a unit test!

royvanrijn: I have some local classes that test; problem is that I believe it works on my machine, just not on the target machine, I’ll check soon.

royvanrijn: Yeah, annoying, it runs fine locally (the code that was pushed), sigh. Kind of debugging in the dark haha... a challenge!

franz1981: Thinking about it twice, I'm wrong at https://github.com/gunnarmorling/1brc/pull/5/#discussion_r1440922218

let's say we have a byte[] data = { 0x01, 0x03 }
And we assume to have a short-based version of SWAR

reading the content of data with (a short) little-endian means: 

0x0301

which have the less significant part at the lower address,
hence the binary hex SWAR result obtained (I'm using the Netty algorithm, but here should be the same) will be

0x8000

and, in order to find out 0x03, we have to use the trailing zeros (here 8 + 7 = 15 -> 15/8 = 1) .

Which means that is fine as it is!

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
ddimtirov: I find that `Math.round((double) x / y) / 10.0` produces the same result as the reference - the important part is to do the division in `double` - addition is guaranteed to be loseless.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -15,65 +15,309 @@
...
-        System.out.print(
-                resultMap.entrySet().stream().sorted(Map.Entry.comparingByKey()).map(Object::toString).collect(Collectors.joining(", ")));
-        System.out.println("}");
+                    // SWAR is faster for ';'
+                    int separatorPointer = findNextSWAR(bb, SEPARATOR_PATTERN, startPointer + 3, limit);
 
-        // System.out.println("Took: " + (System.currentTimeMillis() - before));
+                    // Simple is faster for '\n' (just three options)
+                    int endPointer;
+                    if (bb.get(separatorPointer + 4) == '\n') {
```
ddimtirov, line 118: For my input I get IOOBE here:

```
Exception in thread "main" java.lang.IndexOutOfBoundsException
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:542)
	at java.base/java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:567)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:670)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:927)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at dev.morling.onebrc.CalculateAverage_royvanrijn.run(CalculateAverage_royvanrijn.java:144)
	at dev.morling.onebrc.CalculateAverage_royvanrijn.main(CalculateAverage_royvanrijn.java:92)
Caused by: java.lang.IndexOutOfBoundsException
	at java.base/java.nio.Buffer$1.apply(Buffer.java:757)
	at java.base/java.nio.Buffer$1.apply(Buffer.java:754)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
	at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:98)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.nio.Buffer.checkIndex(Buffer.java:768)
	at java.base/java.nio.DirectByteBuffer.get(DirectByteBuffer.java:358)
	at dev.morling.onebrc.CalculateAverage_royvanrijn.lambda$run$0(CalculateAverage_royvanrijn.java:118)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:960)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:934)
	at java.base/java.util.stream.AbstractTask.compute(AbstractTask.java:327)
	at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:754)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
```

Let me know if you need the file.

royvanrijn: Ah yes please, perhaps there is some other bug that's platform dependent; do share.

ddimtirov: Running on Windows. Will check the new version later (it is quite late here now and starting from tomorrow I won't have access to PC for the weekend).

Let me know if you can suggest how to send you the data file - I started bzipping it and it takes forever, but even part way through, the archive is 2Gb (you can mail me upload coordinates at dimitar.dimitrov at gmail dot com)

royvanrijn: If possible, can you narrow it down? Perhaps run a very small test? Do they all crash, just this one?

AlexanderYastrebov: See related https://github.com/gunnarmorling/1brc/issues/61

We've also added some basic samples within https://github.com/gunnarmorling/1brc/pull/82

PR #10, opened 2024-01-02
Implementation using memory mapped file

bjhara:
Implementation entry using memory mapped file (CalculateAverage_bjhara.java)

On my system (Ryzen 7 5700G, 8 cores, 32 GB RAM) it takes appoximately 17 seconds (the reference implementation takes ~2m17s).

### DIFF
--- a/calculate_average_bjhara.sh
+++ b/calculate_average_bjhara.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


royvanrijn:
This is awesome, learned about memory mapped files from looking at your solution. I've learned from your approach and shoehorned it into mine, which made it even faster. On my MacBook Pro (M2) it now runs in <5 seconds 🚀

gunnarmorling:
I've added the missing license header to the source file and merged. Congrats, you're leading now with 00:38:510 on the official evaluation environment!

twobiers:
Cool, learned about a new API. Will try to integrate it in my branch to get hands on experience. Kudos to you!

javamann:
Really cool approach! Your knowledge of Streams beats the crap out of mine. I really like your splitFileChannel. There is a lot of good concepts that I'm going to steal for a side project of mine.

PR #12, opened 2024-01-02
filiphr first submission

filiphr:
Really interesting and fun challenge @gunnarmorling.

I gave it a go, first implementation using Shenandoah GC (reduces GC pauses significantly) and uses parallel iteration.

My implementation is: [CalculateAverage_filiphr.java]
I used OpenJDK Runtime Environment Zulu21.30+15-CA (build 21.0.1+12-LTS) (21.0.1-zulu)

On Macbook Pro (8 core Intel 64 GB RAM) it runs in +/- 1m40s, the reference implementation runs in +/- 4m9s

### DIFF
--- a/calculate_average_filiphr.sh
+++ b/calculate_average_filiphr.sh
@@ -16,5 +16,7 @@
 #
 
 
+sdk use java 21.0.1-graal
+java -version
 JAVA_OPTS=""
 time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_filiphr
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_filiphr.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_filiphr.java
@@ -47,6 +47,8 @@ import java.util.stream.StreamSupport;
  * When using Oracle GraalVM 21.0.1+12.1
  * `sdk use java 21.0.1-graal`
  * It takes 0m 15s on my machine
+ * `sdk use java 21.0.1-graalce`
+ * It takes 0m 20s on my machine
  *
  * @author Filip Hrisafov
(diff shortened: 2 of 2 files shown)


gunnarmorling:
Nice one, thanks a lot! Will evaluate this one tomorrow, first need to install that Zulu JDK.

filiphr:
Thanks. It doesn't have to be Zulu though. It needs a JDK that has Shenandoah GC, the ones provided by Oracle do not have it (Graal, Oracle JDK and Open JDK). Zulu, Temurin, RedHat, SAP have it (in case it is easier for you to use one of the others)

filiphr:
Hey @lobaorn, it would for sure be interesting to try out a different build from https://builds.shipilev.net/. Which one will you suggest me to try out?

I've been also working on using memory mapped files locally and it seems like the GC is not making any special differences now.

filiphr:
Thanks @lobaorn, I'll check Twitter and the Comments in #15 as well.

I've done some rework on this using memory mapped files thanks @bjhara and @royvanrijn for their submissions, I learned something new. 

Changed to use `BigDecimal` parsing instead of `Double.parseBoolean`, that improves a bit the speed as well.

With the updates it now runs in ~35s on my machine

gunnarmorling:
@filiphr, if any of the distros available from SDKMan can be used, could you add a line [like this](https://github.com/gunnarmorling/1brc/pull/45) to your launch script?

filiphr:
@gunnarmorling, done. I've added `sdk use java 21.0.1-graal`. Using the Graal CE edition makes no difference for me. 

If it isn't possible to use Oracle GraalVM (it is available on SDKMan) then it is fine to just run this with the default JDK

filiphr:
Nice, thanks for merging it. Such a fun experience. Will ty to speed it up a bit more :)

PR #17, opened 2024-01-02
 padreati's first try

padreati:
I have used virtual threads to process in parallel chunks of the input file, some SIMD Vector API to check for new lines and commas, actually nothing too fancy.

On my laptop Ubuntu i7 4 cores, 8 threads I get something around 45 seconds, the original benchmark being 2min51sec.

I have also added some configuration for using vector api and enable preview in the maven pom file.

### DIFF
--- a/calculate_average_padreati.sh
+++ b/calculate_average_padreati.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Woot, woot, first implementation using the Vector API!

gunnarmorling:
Nice one! Clocking in at 0:50:547 in the official evaluation environment (i.e. interestingly, the relative improvement there is quite a bit bigger than in the numbers you shared). Perhaps wider AVX lanes (it's some AMD EPYC chip)?

PR #19, opened 2024-01-02
Implementation - palmr

Palmr:
My implementation: `CalculateAverage_palmr`
Was using `Zulu21.30+15-CA (build 21.0.1+12-LTS)` for JDK but nothing specific to Zulu.

Memory mapped file, single-pass parsing, custom hash map, fixed size thread pool
The threading was a hasty addition and needs work. On a single thread we got it down to ~40s locally.

An assumption has been made that measurement values are all 1 decimal place. Hoping that's fair :crossed_fingers: 

Locally seeing on Linux 6.1.8 (Fedora 36), AMD Ryzen 9 5950X, 128GB RAM, NVMe disk (AKA not representative of the SUT at all, but hey):

```shell
#  ./calculate_average.sh
real    2m3.411s
user    2m1.455s
sys     0m2.821s

# ./calculate_average_bjhara.sh
real    0m10.318s
user    4m22.220s
sys     0m10.747s

#  ./calculate_average_palmr.sh
real    0m5.787s
user    0m42.884s
sys     0m0.796s
```

### DIFF
--- a/calculate_average_palmr.sh
+++ b/calculate_average_palmr.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


Palmr:
> Shamelessly sharing this idea for JVM/GC tuning in another PR/discussion? [#15 (comment)](https://github.com/gunnarmorling/1brc/pull/15#issuecomment-1875495420)

I think the ZGC parts are moot for my solution, since it doesn't GC. I should probably use Epsilon GC but when trying locally it didn't impact the runtime.

I've not seen anything about Project Leyden before, so something to play around with later.
It looks similar to Azul ReadyNow? I did consider trying but avoided since it's paid only.

gunnarmorling:
Nice one, clocking in at 00:18.865. Squashing and merging. Exploring Leyden would be great, it was one of the thing I was hoping someone would do!

PR #21, opened 2024-01-02
Reading and parsing optimizations: CalculateAverage_spullara

spullara:
On my MacBook Pro M3 Ultra:

CalculateAverage: 2m37.788s
CalculateAverage_royvanrijn: 0m29.639s
CalculateAverage_spullara: 0m6.278s

Mostly optimized reading the file and parsing it.


spullara:
I find that the actual runtimes vary quite a bit from run to run. Not sure what is causing that. Happens on both Windows & Mac. Might be something related to file caching.

spullara:
Using Graal, it gets down to ~4.5s.

deemkeen:
@spullara really cool, i even get a slightly better performance with _Executors.newVirtualThreadPerTaskExecutor()_ instead of the fixed threadPool

spullara:
Latest commit slightly beats the performance of the leaderboard leader at 2900ms using GraalVM. FileChannel was the key to closing the gap. Have to see if improving the hash map helps a lot.

spullara:
Now down to 2s after writing my own hash map.

gunnarmorling:
Very nice, clocking in at 00:14.848 in the evaluation environment!

PR #29, opened 2024-01-03
Add initial implementation by seijikun

seijikun:
Thanks for the interesting challenge!
I'm not too experienced with Java but I had a go at it anyway, without looking at any of the other submissions.

My implementation uses memory mapped files and chunks the measurement file into as many chunks as the host processor has threads.
I also used a custom double parser method, since a fully compliant one is expensive.

On my notebook, I have a runtime of 13 - 14 seconds


gunnarmorling:
Nice, 38.106 sec on the eval env!

PR #31, opened 2024-01-03
Usage of try-with-resources

khmarbaise:
pom file cleanup


gunnarmorling:
Nice, thanks for the house keeping!

PR #32, opened 2024-01-03
ddimtirov - fixpoint, objects, manual framing and parsing

ddimtirov:
Please run with the Oracle JDK 21.

On my machine, `CalculateAverage`  takes `02:10` to run. HEre are the changes I've made so far and the corresponding times:

1. **Iteration 1 (0:44):** switching to fixed point integer arithmetic, simpler integer parsing, fewer allocations, parallel streams, and parallel GC
2. **Iteration 2 (0:37):** single-thereaded, manual framing in I/O using `Reader` and `CharBuffer`, processing (one pass, no allocation, no copying)
3. **Iteration 3 (0:02.3):** Same processing over mmapped parallel i/o
4. **Iteration 4 (0:02.1):** Switched from byte buffers to the `MemorySegment` preview API - thread confined and slightly faster; while at it also disabled GC

Rebased and ready for merge.

Rejected options: 
* Dynamic AppCDS - adds more overhead than benefit
* Native Image - 50% slower

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_ddimtirov.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_ddimtirov.java
@@ -15,14 +15,18 @@
  */
 package dev.morling.onebrc;
 
+
 import java.io.*;
-import java.nio.CharBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Map;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.TreeMap;
 
 // gunnar morling - 2:10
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Nice! Taking the lead with 00:11:886 on the eval environment. Great to see usage of the new foreign memory API!

src/main/java/dev/morling/onebrc/CalculateAverage_ddimtirov.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.TreeMap;
+
+// gunnar morling - 2:10
+// roy van rijn -   1:01
+//                  0:37
+
+public class CalculateAverage_ddimtirov {
+    private static final String FILE = "./measurements.txt";
+
+    private static final int HASH_NO_CLASH_MODULUS = 49999;
```
gunnarmorling, line 40: I might have been a bit too quick merging this one. How is that value determined, does it happen to prevent hash collisions just for the set of station names currently showing up in the data set? I.e. is the solution guaranteed to work also for other possible station names? If not, I'm afraid this cannot be accepted as is, as any solution must work for all names:

> A: No, while only a fixed set of station names is used by the data set generator, any solution should work with arbitrary UTF-8 station names (for the sake of simplicity, names are guaranteed to contain no ; character).

gunnarmorling:
Hey @ddimtirov, just realizing that there may be an optimization at play here which isn't considered valid. See my comment above.

src/main/java/dev/morling/onebrc/CalculateAverage_ddimtirov.java
sundy-li: This may work for this dataset, but not for other large datasets where hash collisions are common.

ddimtirov:
Hi @gunnarmorling - apologies, I had missed that clause. 

Yes, the value is selected offline, so please take this off the leaderboard for now. I'll think of another data structure that can handle arbitrary names.

BTW, I see that the current leading submission has the same problem, and likely sime other too: https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java#L30, 
https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java#L149-L155

PR #34, opened 2024-01-03
padreati's second try

padreati:
Some small improvements on double parsing, parse lengths, no offsets, etc.

On my host the running time is 23sec, first try was 45sec, baseline was 2min50sec.


padreati:
I have a new version, which is much better. I decided to close that.

Issue #14, opened 2024-01-02
Use CSV output format


AlexanderYastrebov:
Use proper CSV (or semicolon-separated for that matter) output format
```csv
Addis Ababa;33.0;33.0;33.0
Aden;31.1;31.1;31.1,
...
```
instead of weird not-a-json oneliner output.

This would simplify result comparison (diff) between multiple implementations.
This would also enable parsing of results - think of import into database or e.g. database-based implementation.

gunnarmorling:
I agree that this would have been the better way. I'm hesitating though to change the format at this point, to keep different implementations comparable.

lluismf:
The weird not-a-json oneliner is just the map serialized. Writing a CSV output would be extra cost.

AlexanderYastrebov:
@lluismf You are right, serializing result in a portable format is a game-changer and will severely affect processing performance of 10^9 input rows :+1:

gunnarmorling:
Hey, let's keep it friendly :)

I agree that a different output format wouldn't make any difference perf-wise in the grand scheme of things and as said, it would have been the better choice. But I don't think it's that much of an issue to justify changing this while the challenge is running.

So I'd keep this as-is for the time being, and consider it a lesson learned for whenever another challenge of this kind is happening. Thanks all!

PR #36, opened 2024-01-03
Add a script to transform output into CSV format

AlexanderYastrebov:
It is useful to debug differences between implementations, e.g.:
```sh
$ while ./create_measurements.sh 1000 && diff <(./calculate_average_royvanrijn.sh 2>/dev/null | ./tocsv.sh) <(./calculate_average.sh 2>/dev/null | ./tocsv.sh) ; do echo OK; done
Created file with 1,000 measurements in 50 ms
60c60
< Bucharest;-2.9;2.9;6.1
---
> Bucharest;-2.9;2.8;6.1
265c265
< Petropavlovsk-Kamchatsky;0.9;9.3;17.7
---
> Petropavlovsk-Kamchatsky;0.9;9.2;17.7
```

For #14

### DIFF
--- a/tocsv.sh
+++ b/tocsv.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Nice one, thanks!

PR #37, opened 2024-01-03
Experiment from entangled90

entangled90:


gunnarmorling:
Could you please rebase this one, so as to resolve the merge conflict and then remove the "needs rebase" label again? Also please make sure that running `./test.sh entangled90` passes. Thex

entangled90:
Done, sorry about that!

gunnarmorling:
Hey @entangled90, tests pass now, but the output is different from that one of the baseline implementation.

entangled90:
There was an error with rounding. Thanks for catching it!

royvanrijn:
I've found an issue in this entry, if we add the following line in the print results:

        System.out.println("Processed: " + processors.entrySet().stream().mapToLong(e -> e.getValue().count).sum());

We can see that it is only processing 312843874 records instead of a billion, it seems to miss about 2/3rds of the entries for some reason. Can't really pinpoint where the issue is though (yet).

royvanrijn:
Ah, it has to do with the size of the queue and the MAX_MAPPED_MEMORY. If I make both `1*1024*1024` in size, we get `1000000000` entries.

entangled90:
Thanks for spotting this! Issue is I was using  ArrayBlockingQueue.offer (which does not block if queue is full). On my pc there is no difference in time so probably depends on the machine/core available

gunnarmorling:
00:18.380, nice. Thx for particpating!

PR #55, opened 2024-01-03
Small cleanups

filiphr:
This PR does a cleanup and a small fix for IntelliJ:

* There is an auto reformat happening during the build, so it would be good to have no changes when running locally for everyone
* I know that @khmarbaise did an improvement for the maven-compiler-plugin in #31. However, IntelliJ does not yet understand the `enablePreview` configuration option of the plugin (see [IDEA-296303](https://youtrack.jetbrains.com/issue/IDEA-296303)). Without this I can't run my class in IntelliJ as it complains about using preview features

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,9 +101,14 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.12.1</version>
           <configuration>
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Good stuff, thanks, @filiphr!

PR #65, opened 2024-01-03
filiphr second submission

filiphr:
Did some profiling and tried to remove some additional hotspots:

* Use long instead of double parsing
* Use an integer hash key for the intermediate map to avoid creating strings that won't be used
* Manually compute the value instead of using `Long#parseLong`

With this I have managed to get to ~7s on my machine using Oracle GraalVM


gunnarmorling:
Nice! 00:12.270 now.

gunnarmorling:
Hey @filiphr, as discussed with @spullara I had to remove your entry for the time being, as it's prone to hash collisions. Happy to add it back when that issue has been addressed.

PR #66, opened 2024-01-03
padreati's third submission

padreati:
I managed to reduce time on local machine to 13 seconds. Some notes regarding what I have done:

- no memory mapped files
- split file processing in chunks
- use vector api in a better way than previously
- create a shared name registry with synchronization to avoid creating new string objects and work on hashes
- various other improvements

On my local laptop i7 4 cores, 8 threads, AVX512 it takes ~13 seconds. I expect to be slower, around 15 secs on benchmark machine.


gunnarmorling:
Could you rebase this one to the latest main branch and ideally squash everything into a single commit too? And then remove the "needs rebase" label? Thanks!

PR #69, opened 2024-01-03
Implementation jgrateron

jgrateron:
execution in 00:01:17 Ryzen 5/6 core 12 threads/ 16GB Ram


gunnarmorling:
01:27.912. Thanks for participating!

PR #70, opened 2024-01-04
Initial version for thomaswue with Oracle GraalVM Native Image

thomaswue:
The script "additional_build_step_thomaswue.sh" will generate the image as "image_calculateaverage_thomaswue". The script "calculate_average_thomaswue.sh" will either execute the native image if the file exists or otherwise run the program in JVM mode with the Graal JIT compiler.

The program finishes on my system in 1.67s (total CPU user time is 42.6s). It is using an Intel 13th Gen Core i9-13900K processor.

**Update:** Thanks to tuning from @mukel using sun.misc.Unsafe to directly access the mapped memory, it is now down to 1.28s (total CPU user time 32.2s) on my machine. Also, instead of PGO, this is now just using a single native image build run with tuning flags "-O3" and "-march=native".


thomaswue:
You are certainly very welcome to copy the GraalVM native image generation scripts and see if it helps also with your solution! Key is to use profile-guided optimizations and to have an already optimized solution that does not run for too long. I don't think Lilliput or Valhalla (or Loom) can be beneficial for this benchmark.

mariusstaicu:
Was thinking of trying this too combined with one of the top implementations.
Also, would be a lot of fun comparing same algo with different jdks with or without native image.

thomaswue:
Agreed. And there possibly also different input sizes to more isolated show the startup vs long term peak performance characteristics.

calculate_average_thomaswue.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [ -z "${JVM_MODE}" ]; then
+    echo "Chosing native image mode, set JVM_MODE variable to select JVM mode."
+    if [ -f ./image_calculateaverage_thomaswue ]; then
+        echo 'Picking up existing native image.'
+    else
+        if [ -f ./profile_thomaswue.iprof ]; then
+            echo 'Picking up profiling information from thomaswue.iprof.'
+        else
+            echo 'Could not find profiling information, therefore it will be now regenerated.'
```
gunnarmorling, line 31: Hey @thomaswue, thanks a lot for this, very interesting to see PGO!

In its current shape, I think though that it goes somewhat against the spirit of this challenge, specifically "Implementations must not rely on specifics of a given data set". Which I think we kinda do when leveraging profile data from the first run during following ones. Admittedly, the wording could be more precise, but the general spirit is that runs shouldn't benefit from outcomes from previous runs (as otherwise, when taken to the extreme, one could calculate results once store them in a file and then simply load the content of that file in the subsequent runs).

What would be acceptable to me though is creating the profile data at build time, not relying on the specific data set used during evaluation (similar to how creating a static CDS archive is in bounds). Would that be an option?

thomaswue: Sure, absolutely. The profiling is only used to automate compiler tuning knowledge one would otherwise need to manually specify (e.g. "inline a lot in this method"). I can for example train on the existing test data files or check in a slightly larger test file with maybe 1k lines?

Another thing maybe in the spirit of the competition to generate interesting insights could be to run that same algorithm in 3 different ways - i.e. JIT, AOT, AOT+PGO.

gunnarmorling: >  The profiling is only used to automate compiler tuning knowledge one would otherwise need to manually specify (e.g. "inline a lot in this method")

Yepp, that's what I kinda thought.

> check in a slightly larger test file with maybe 1k lines?

Yeah, if you could add a 1K file and the expected output (as obtained via _calculate_average_baseline.sh_) to _src/test/resources_ that would be perfect. It would extend the tests and you can use it for "training". Can the PGO data be generated via Maven? If so, that'd be best, but it's not a strict requirement. A one-off script to run before the evaluation would be ok, too.

> Another thing maybe in the spirit of the competition to generate interesting insights could be to run that same algorithm in 3 different ways - i.e. JIT, AOT, AOT+PGO.

Agreed. Thing is, I'm really overwhelmed by the number of submissions and I won't have any capacity for further increasing scope of this right now. But you're more than welcome to do this experiment and publish any interesting insights on your own end.

thomaswue: OK, sounds good. As a quick fix, I have turned off PGO by default and instead added global tuning flags in a commit added to this PR. Maybe approaches that require running the application before should be kept separate in terms of evaluation anyway.

Agreed that running with different configs probably increases the scope too much. It could be interesting to do this after 31st of Jan for a few of the submissions.

Let me know if we can assist you with some of the work. Obviously a downside of the success of the project to have so much attention ;-).

gunnarmorling: Ok, excellent, will evaluate this one tomorrow (need to install the native image tool first). In the mean time, could you make sure that running this one shows no differences to the expected output:

```
./test.sh thomaswue
```

We've added this one just earlier today so as to have at least some basic assurance of correctness for implementations. Thanks again!

thomaswue: OK, will check. You do not need to install the native image tool, because it is included for the latest Graal versions. So just "sdk install java 21.0.1-graal" is enough.

thomaswue: Had to make the script more quiet and fix a minor issue when accumulating results, but now "./test.sh thomaswue" succeeds (two new commits added). The script will generate on first invocation the image with some extra output though. Let me know if this is appropriate or there should be a different way to integrate this into the build process.

imrafaelmerino: I think to be honest that all that time cooking the image should be added to the overall result.

imrafaelmerino: Don't get me wrong. I think is great to learn new things and put Java to the test. Just an opinion. From what I'm seeing (not related to this) Java 21 is the new C! tailor made hashes and data structures, branchless programming, varhandles etc. I think it would be better a simple and idiomatic solution. Anyways, waiting for your solution and sure I'll take a look to learn this approach, have a nice weekend and keep it up!

thomaswue: Yes, I think it would be interesting to have some categories such as most idiomatic solution or solution with smallest amount of statements. Then one can test different JVM implementations and how they fare for each. When asking only for the fastest, naturally one gets a solution that might look very close to C code. However, this at least shows imho that if necessary Java can also be used on such low level.

gunnarmorling: Yeah, I can see that angle, OTOH, it's super-hard to impossible to objectively rate what's "most idiomatic". Probably something better suited for a blog post or a talk where one can explore these nuances. That said, I think there's somewhere an interesting threshold of how far comparatively "standard" Java solutions go up on the leaderboard (quite far in fact) before the more extreme solutions kick in. I.e. "idiomatic" gets you surprisingly far.

gunnarmorling: > The script will generate on first invocation the image with some extra output though. The image file would have to be cleaned up when updating the code (e.g., after the evaluation). Let me know if this is appropriate or there should be a different way to integrate this into the build process.

@thomaswue, so IIUC, this still does the profiling with the actual dataset in the first run, right? Didn't you mean to provide a separate dataset file and use this one when creating the native image? I.e. the build (with profiling) should be clearly be separated from the evaluation run. Can we extract the part which builds the native image (including PGO) to a separate script and invoke this one to build the application (instead of the usual `mvn verify`)? 

So the flow would become this:

Build: 
```
ln -s src/test/data/yourprofilingdataset.txt measurements.txt
./your_native_build_script.sh # builds the native binary and PGO data
```

Evaluation (i.e. what I am doing):

```
rm -f measurements.txt
ln -s measurements_1B.txt measurements.txt

for i in {1..5}
do
    ./calculate_average_thomaswue.sh # just the actual launch command, no build steps
done
.
```

thomaswue: The use of PGO is disabled (just had the option to enable via environment variable). To avoid confusion and make this simpler, I deleted the code for optionally using PGO for now. There is an "additional_build_step_thomaswue.sh" that builds the image. The "calculcate_average_thomaswue.sh" script runs in JVM mode if the file is not there and otherwise picks up the generated image. Let me know if this now works better for you.

thomaswue: I have also updated the PR description to clarify the new behavior and that PGO is not in use for this version.

gunnarmorling: Yepp, this LGTM now, thanks!

gunnarmorling:
@thomaswue, very cool. Coming in at 9.625sec on the (8 core) eval machine, i.e. 2nd place! Would love to see the PGO version as a follow-up, created with that additional build script you've provided. Thanks for participating!

gunnarmorling:
Squashed and merged. Thx!

gunnarmorling:
> Maybe approaches that require running the application before should be kept separate in terms of evaluation anyway.

I have added a "Note" column to the leaderboard, stating for this entry that this is a GraalVM native binary.

> Agreed that running with different configs probably increases the scope too much. It could be interesting to do this after 31st of Jan for a few of the submissions.
>
> Let me know if we can assist you with some of the work. Obviously a downside of the success of the project to have so much attention ;-).

Thanks :) I might take you up on this at some point. I think once the dust has settled a bit, we can explore how to make use of this set-up for running all different kinds of comparisons.

thomaswue:
Cool, thank you for merging and evaluating!

Yes, making comparisons for a few solutions with different run options (and maybe also input sizes or target hardware) is certainly something I would be interested in doing.

PR #71, opened 2024-01-04
Add CalculateAverage_moysesb

moysesb:
Yet another mmap based implementation. My implementation cannot compete with the ones using open hash arrays, but I'm curious to see how it does in the target machine.

My system has a i7-8750H CPU (6 cores ,12 "threads") and 16GB RAM. My implementation finishes in ~22s, while the reference implementation finishes in ~3.2min.

### DIFF
--- a/calculate_average_moysesb.sh
+++ b/calculate_average_moysesb.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Could you please take a look at the test failures reported by

```
./test.sh moysesb
```

And remove the "test failure" label once done? Thx!

moysesb:
Oops, sorry, should have run the tests yesterday before pushing.

moysesb:
About the test failure label, I don't think I have write access to labels in this repo.

gunnarmorling:
> Oops, sorry, should have run the tests yesterday before pushing.

No worries, just added this earlier today to cover some corner cases (should have done so from the beginning).

gunnarmorling:
@moysesb, something still is wrong, the output differs from that of the reference implementation (min/max values). Could you take a look?

moysesb:
What's the procedure for checking the calculated values? I generated a couple of measurement inputs locally with create_measurements.sh (from 1_000 to 10_000_000 lines) and the results are always the same between my class and the baseline.

gunnarmorling:
Can you try and create one with 1B rows actually? If I'd have to guess, I'd
say there's some value overflow somewhere.

moysesb:
Yeah, that was it. Last chunk was being truncated because I can't do basic arithmetic :/

Fixed (I hope).

gunnarmorling:
00:32.764 Nice!

PR #74, opened 2024-01-04
Use proper hash key collision detection scheme.

ebarlas:
This change is primarily about compliance. The prior version of my `ebarlas` submission didn't adequately handle theoretical hash key collisions. The current version now does that with an open addressing scheme.

Changes here also include various unrelated adjustments.

Implementation class: `CalculateAverage_ebarlas`

The exec time on an AWS EC2 `c5.2xlarge` with 8 vCPU is slightly higher than before, now clocking in at 10.1s. However, still noticeably lower than on the CX33 target env, interestingly.

```
real	0m10.192s
user	1m19.001s
sys	0m0.580s
```

This submission is still configured to run on GraalVM CE.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java
@@ -51,7 +51,8 @@ public class CalculateAverage_ebarlas {
                 try {
                     var buffer = channel.map(FileChannel.MapMode.READ_ONLY, pStart, pSize);
                     partitions[pIdx] = processBuffer(buffer, pIdx == 0);
-                } catch (IOException e) {
+                }
+                catch (IOException e) {
                     throw new RuntimeException(e);
                 }
             };
@@ -94,7 +95,8 @@ public class CalculateAverage_ebarlas {
                     var t = target[j];
                     if (t == null) {
                         target[j] = current[j]; // copy ref from current to target
-                    } else {
+                    }
+                    else {
                         t.min = Math.min(t.min, current[j].min);
                         t.max = Math.max(t.max, current[j].max);
                         t.sum += current[j].sum;
@@ -154,10 +156,12 @@ public class CalculateAverage_ebarlas {
                         st = stats[idx] = new Stats(key);
                     }
                     readingKey = false;
-                } else {
+                }
+                else {
                     keyHash = HASH_FACTOR * keyHash + b;
                 }
-            } else {
+            }
+            else {
                 if (b == '\n') {
                     var v = negative ? -val : val;
                     st.min = Math.min(st.min, v);
@@ -169,9 +173,11 @@ public class CalculateAverage_ebarlas {
                     val = 0;
                     negative = false;
                     keyStart = buffer.position();
-                } else if (b == '-') {
+                }
+                else if (b == '-') {
                     negative = true;
-                } else if (b != '.') { // skip '.' since fractional tenth unit after decimal point is assumed
+                }
+                else if (b != '.') { // skip '.' since fractional tenth unit after decimal point is assumed
                     val = val * 10 + (b - '0');
                 }
             }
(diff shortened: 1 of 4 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java
```diff
@@ -21,13 +21,15 @@
...
 import java.util.List;
 import java.util.TreeMap;
 
 public class CalculateAverage_ebarlas {
 
-    private static final int HASH_FACTOR = 278;
-    private static final int HASH_MOD = 3_487;
+    private static final int MAX_KEY_SIZE = 64;
+    private static final int HASH_FACTOR = 433;
+    private static final int HASH_TBL_SIZE = 16_383; // range of allowed hash values, inclusive
```
ddimtirov, line 32: As the table is not growing, this assumes there will be no more than 16k stations.

This is not acceptable (see https://github.com/gunnarmorling/1brc/discussions/18#discussioncomment-8004520 and the remainder of the thread).

ddimtirov, line 32: As the backing array cannot be grown, this hash table will not be able to handle more than 16k entries.

This again violates the spirit of not making assumptions about the exact values of the data (see this thread https://github.com/gunnarmorling/1brc/discussions/18#discussioncomment-8004520).

And yes, I fell for it too :-)

ebarlas: Thanks for pointing that out! However, this Twitter exchange seems to suggest otherwise:

https://twitter.com/royvanrijn/status/1742878555707519363?t=8KIWFDbq6l0mpdgj_BxGhw&s=19

ebarlas: Screenshot_20240104-060031

gunnarmorling: Yepp, I think it makes sense to limit the maximum number of distinct stations to keep the effort for submissions at check. I've just clarified this in the README, too.

ebarlas: Let me know if you have any other concerns about my submission, @gunnarmorling. I believe it is in compliance.

ddimtirov: Thanks for letting me know about this - now I feel good for not finding time to implement growing and rehashing ;-)

gunnarmorling: > Let me know if you have any other concerns about my submission, @gunnarmorling. I believe it is in compliance.

Awesome, will do. Working on automating the evaluation process right now (it's just too much for doing it manually) and also adding a little test suite. So it will take a little while until I can continue evaluating, but it should hopefully be faster and more thorough and reliable then.

ebarlas: Sounds good. I just glanced at the open PRs - there are a lot of them!

I also happened to notice at least one PR which is a copy-paste of my entire prior solution (now out of compliance).

I wonder if it makes sense to add a notice about original work. Otherwise the leaderboard (which is a fun motivator) is sort of meaningless.

ddimtirov: If it is plain copy/paste it would perform the same, so nothing lost. Perhaps it is also worth declaring ties if times are within 50ms from each other (the range of error).

BTW, looking at your solution, it was strikingly similar to mine (though I used more arrays , you used a data class), so perhaps there are only so many ways to skin that cat ;-)

gunnarmorling: This situation already is discussed in the FAQ:

Screenshot 2024-01-04 at 16-42-09 gunnarmorling_1brc 1️⃣🐝🏎️ The One Billion Row Challenge -- A fun exploration of how quickly 1B rows from a text file can be aggregated with Java

I.e. copying in general is ok, but there should be meaningful progress made in addition. If you think there is a submission which goes against that, can you tag me on that PR, ideally with the output of `git diff` between the two files? Thanks!

PR #75, opened 2024-01-04
Submitting my version

obourgain:
My implementation is in dev.morling.onebrc.CalculateAverage_obourgain and runnable with provided script calculate_average_obourgain.sh. Runs with standard JDK 21.

On my computers (i5 13500, 20 cores, 32GB ram) my best run is (file fully in page cache):
39.26user 0.67system 0:02.24elapsed 1780%CPU

As I wrote in comments in the code, I have a few different roundings that the reference implementation. I have seend that there is an issue about that, but no specific rule yet.

Main points:
- use MemorySegment, it's faster than ByteBuffer
- split the work in a lot of chunks and distribute to a ForkJoinPool
- very low allocation (runs with Xmx24m on my computer)
- visit each byte only once, hash as you go

Things I tried that were in fact pessimizations or useless:
- use some internal JDK code to vectorize the hashCode computation
- use a MemorySegment to represent the keys instead of byte[], to avoid copying
- SWAR for finding the semicolon
- SWAR to subtract '0' from all the temperature bytes at once

Things that could be done:
- reading with long instead of ints. That may make SWAR to find the semicolon worth it, but I would have to support the case were two semicolons are in the same long, and I already spent too much time on this.
- more vectorization?

My takeaways:
- simple code can already get very fast
- the JVM has good defaults
- MemorySegment API is nice
- I had to spend a lot of time to get small gains, and they often made the code a lot more complex
- learnt a ton while coding and reading other submissions

Hope I won't have a bad surprise when running on the target server 😱

### DIFF
--- a/calculate_average_obourgain.sh
+++ b/calculate_average_obourgain.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# runs with -Xmx24m on my machine, playing it safe with a larger heap
+JAVA_OPTS="-Xmx64m --enable-preview"
+# to use some black magic options
+JAVA_OPTS="$JAVA_OPTS -XX:+UnlockExperimentalVMOptions"
+# no GC, not needed
+JAVA_OPTS="$JAVA_OPTS -XX:+UseEpsilonGC -XX:+AlwaysPreTouch"
+# my finals are really final
+JAVA_OPTS="$JAVA_OPTS -XX:+TrustFinalNonStaticFields"
+# to get CalculateAverage_obourgain$OpenAddressingMap::getOrCreate to inline. A compile command wasn't enough, it was still hitting 'already compiled into a big method'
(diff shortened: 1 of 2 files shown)


gunnarmorling:
This fails one of the tests:

```
./tests.sh obourgain
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-10000-unique-keys.txt

real	0m0.257s
user	0m0.618s
sys	0m0.079s
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-10.txt

real	0m0.110s
user	0m0.099s
sys	0m0.038s
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-1.txt

real	0m0.126s
user	0m0.094s
sys	0m0.061s
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-20.txt

real	0m0.128s
user	0m0.106s
sys	0m0.052s
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-2.txt

real	0m0.120s
user	0m0.107s
sys	0m0.043s
Validating calculate_average_obourgain.sh -- src/test/resources/samples/measurements-3.txt

real	0m0.130s
user	0m0.113s
sys	0m0.047s
1c1
< {Bosaso=-15.0/1.2/20.0, Petropavlovsk-Kamchatsky=-9.5/0.0/9.5}
---
> {Bosaso=-15.0/1.3/20.0, Petropavlovsk-Kamchatsky=-9.5/0.0/9.5}
```

Could you take a look?

hastebrot:
@obourgain your version with `BitTwiddledMap` was already very fast. I wonder what resulted in the tests to fail, and how a fix would influence the performance.

obourgain:
I pushed a fixed version.
All test pass.
@hastebrot the test failure was due to a difference in rounding, no measurable difference in rounding.
The BitTwiddledMap was from a previous submission from royvanrijn but didn't support collision. I now have a more generic version.

obourgain:
@gunnarmorling my submission passes the tests now, could you run it again please?

gunnarmorling:
00:09.062, nice one! Currently on #3. Thx for particpating!

PR #52, opened 2024-01-03
doc: document flamegraph generation with jbang

maxandersen:

PR #78, opened 2024-01-04
Link async-profiler and ap-loader in the README

parttimenerd:
As an addendum to #52, adding links to async-profiler and ap-loader.


gunnarmorling:
Very thorough :) Merging.

Issue #61, opened 2024-01-03
Need acceptance test suite


AlexanderYastrebov:
E.g. current "winner" produces garbage output on trivial input:
```
$ ./create_measurements.sh 2
Created file with 2 measurements in 18 ms
$ cat measurements.txt 
Calgary;1.6
Tamanrasset;13.7
$ ./calculate_average_ebarlas.sh 
{=2.147483647E8/0.0/-2.147483648E8, manrasset=13.7/13.7/13.7, y=2.147483647E8/0.0/-2.147483648E8}

real    0m0,206s
user    0m0,415s
sys     0m0,048s
```

AlexanderYastrebov:
@ebarlas FYI

AlexanderYastrebov:
Something like
```sh
$ for n in 1 2 5 10 100 1000 10000; do ./create_measurements.sh $n && diff <(./calculate_average_AlexanderYastrebov.sh 2>/dev/null | ./tocsv.sh) <(./calculate_average.sh 2>/dev/null | ./tocsv.sh) && echo "OK $n" || echo "FAIL $n"; done

Created file with 1 measurements in 19 ms
OK 1
Created file with 2 measurements in 17 ms
OK 2
Created file with 5 measurements in 21 ms
OK 5
Created file with 10 measurements in 18 ms
OK 10
Created file with 100 measurements in 22 ms
OK 100
Created file with 1,000 measurements in 29 ms
107c107
< Dolisie;5.1;22.5;39.5
---
> Dolisie;5.1;22.6;39.5
124c124
< Gagnoa;17.7;22.4;25.4
---
> Gagnoa;17.7;22.3;25.4
126c126
< Garissa;21.6;41.5;72.0
---
> Garissa;21.6;41.4;72.0
135c135
< Hamburg;-0.9;9.6;17.9
---
> Hamburg;-0.9;9.5;17.9
FAIL 1000
Created file with 10,000 measurements in 62 ms
369c369
< Timbuktu;10.6;29.9;53.7
---
> Timbuktu;10.6;30.0;53.7
FAIL 10000
```
could be used to verify a given implementation.

ddimtirov:
I manually run the reference impl on my generated file and pipe it to a file. Then in my code I have:

```
        assert Files.readAllLines(Path.of("measurements_result.txt")).getFirst().equals(result);
```

It is crude, but it gets the job done... also I find assertions useful alternative to defensive coding when you cannot aford the extra checks on a normal run.

[...]

PR #82, opened 2024-01-04
test: add test samples

AlexanderYastrebov:
Adds test samples that can be used for unit tests or to verify implementations via:
```bash
for sample in $(ls src/test/resources/samples/*.txt)
do
  echo "Validating $sample"
  rm -f measurements.txt
  ln -s $sample measurements.txt

  diff <(./calculate_average.sh) ${sample%.txt}.out
done
rm measurements.txt
```

For #61


AlexanderYastrebov:
This is just an initial set of trivial imputs, other samples can be added later, e.g. that do not contain station names used by the reference implementation.

AlexanderYastrebov:
Ideally each implementation should support a way to specify an input file, e.g. via commandline arg or env variable like
```java
private static final String FILE = Optional.ofNullable(System.getenv("MEASUREMENTS_FILE")).orElse("./measurements.txt");
```

gunnarmorling:
> Ideally each implementation should support a way to specify an input file, e.g. via commandline arg or env variable like...

Yeah, that would have been nice, but I'd leave that for another time. I'd like to minimize impact on existing submissions as much as possible, in particular if it's something we can work around, as with linking files here as you did.

I'll take care of the build failure and merge it. Will also put your loop above and put it into a script. Thanks a lot for pushing this, I think it's a great start!

PR #84, opened 2024-01-04
test: Add samples unit test

AlexanderYastrebov:
This is just a fun excersise, the real validator should use shell script like in #82 since many implementations depend on different jdks and command line arguments.

For #61

### DIFF
--- a/src/test/java/dev/morling/onebrc/TestCalculateAverage.java
+++ b/src/test/java/dev/morling/onebrc/TestCalculateAverage.java
@@ -1,0 +1,128 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


filiphr:
Cool idea @AlexanderYastrebov. An alternative for something like this might also be to use [Dynamic Tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests) with a `@TestFactory`

The classes could be dynamically found. e.g. using [CLassGraph](https://github.com/classgraph/classgraph)

adding dependency

```xml
    <dependency>
      <groupId>io.github.classgraph</groupId>
      <artifactId>classgraph</artifactId>
      <version>4.8.165</version>
      <scope>test</scope>
    </dependency>
```

We can have a test that does something like:

```
    @TestFactory
    Stream<DynamicTest> testSamples() {
        String pkg = "dev.morling.onebrc";
        try (ScanResult scanResult =
                     new ClassGraph()
                             .enableClassInfo()
                             .acceptPackages(pkg)     // Scan com.xyz and subpackages (omit to scan all packages)
                             .scan()) {               // Start the scan
            return scanResult.getAllClasses()
                    .stream()
                    .filter(classInfo -> classInfo.getSimpleName().startsWith("CalculateAverage"))
                    .map(classInfo -> DynamicTest.dynamicTest(classInfo.getSimpleName(),
                            () -> TestSamples(TestCalculateAverage.class.getClassLoader().loadClass(classInfo.getName()))));
        }
    }
```

`() -> TestSamples(TestCalculateAverage.class.getClassLoader().loadClass(classInfo.getName())))` is calling the test method that is already in there.

Side note, using something like a `@TestFactory` we could even lookup the `calculate_average_<username>.sh` files and invoke `test.sh`.

gunnarmorling:
Do we still need this actually? Feel like the current shell test set-up is good enough for what we need?

PR #89, opened 2024-01-04
Unit test calculate average test

radughiorma:
I added a unit test that uses the sample files you provided. 
I had to the change the submitted classes to read the input file from the arguments passed so that.

The test is not passing for all submissions.


AlexanderYastrebov:
See also https://github.com/gunnarmorling/1brc/pull/84

gunnarmorling:
This one is modifying way too many files. It should only add your implementation and its launch scripts. I am going to close this one, so as to keep the slate clean. When reopening, please make sure to have a single commit off of latest main and the the ./test.sh script passes for your implementation. Thx!

PR #97, opened 2024-01-04
test: add sample with 10k unique keys

### LINKED ISSUE #91: Add test with 10K distinct keys to the suite
Otherwise, we won't be able to detect hash collisions, it's too easy to miss during review.

AlexanderYastrebov:
Input created via
```sh
bash -c 'for i in {1..10000} ; do echo "id$i;0.0" ; done' >./src/test/resources/samples/measurements-10000-unique-keys.txt
```
and output via baseline implementation.

Keys are short and very similar which improves chances for collision
and hence are good for testing.

Fixes #91


.gitignore
```diff
@@ -43,5 +43,5 @@ out/
 
 # 1BRC
 /measurements*.txt
-*.out
+/*.out
```
AlexanderYastrebov, line 46: Not sure why this was ignored, changed to ignore only in the repo root

AlexanderYastrebov:
Decided not to bother with supporting gzipped inputs/outputs:
```
$ du -h ./src/test/resources/samples/measurements-10000-unique-keys.*
196K    ./src/test/resources/samples/measurements-10000-unique-keys.out
108K    ./src/test/resources/samples/measurements-10000-unique-keys.txt
```
300k should not be a big deal.

AlexanderYastrebov:
`palmr` hangs and `richardstartin` fails on this input.

.gitignore
gunnarmorling: I use these in the eval tooling to verify the expected output. Just ignoring in the root is fine.

gunnarmorling:
> `palmr` hangs and `richardstartin` fails on this input.

Already logged https://github.com/gunnarmorling/1brc/issues/100 for @Palmr. Will log another one for Richard. Thanks!

PR #98, opened 2024-01-04
Use proper key for CalculateAverage_filiphr

filiphr:
@gunnarmorling I think that with this hash key collisions are being avoided.

@spullara can you please verify. I don't want to make a mistake again 😄.

With this change on my machine it takes ~7.2s


filiphr:
@gunnarmorling thanks for the heads up. Just pushed a commit with that change. Feel free to squash all of it

gunnarmorling:
00:13.857. Key handling LGTM now too. // CC @spullara

PR #107, opened 2024-01-04
Add davecom Entry

davecom:
My main class is `CalculateAverage_davecom`

My implementation uses memory mapped files and runs between 19 and 24 seconds on my M1 MacBook Air (8 cores) with 16 GB of RAM.


gunnarmorling:
Thanks for participating! Could you please run `./test.sh davecom` and verify that there are no test failures before we evaluate (it's gonna be part of the CI build eventually)? Thx!

davecom:
Yes, I did run it before submitting, @gunnarmorling. Thanks!

gunnarmorling:
00:24.879. Thanks for participating!

davecom:
Thanks for creating the contest!

PR #108, opened 2024-01-04
make aggregation state grow dynamically

richardstartin:
This should fix the test failure relating to the hardcoded aggregation state size


gunnarmorling:
All LGTM now. Thx!

PR #110, opened 2024-01-04
test: use temperature value of 1.0

AlexanderYastrebov:
In case of key collision broken implementation will likely attribute measurements to the wrong key and therefore it is better to have non-zero value to end up with a wrong average value.

When all measurements are zero then averages are also zero even when attributed to the wrong keys.

Updates #91


gunnarmorling:
Thx!

PR #114, opened 2024-01-04
merykitty's attempt

merykitty:
My attempt at this challenge, `test.sh merykitty` passed.

Please kindly review, thanks very much.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
@@ -215,7 +215,8 @@ public class CalculateAverage_merykitty {
 
         long currOffset = offset;
         while (true) {
-            long semicolonPos = Long.numberOfTrailingZeros(semicolons);
+            int localOffset = (int) (currOffset - offset);
+            long semicolonPos = Long.numberOfTrailingZeros(semicolons) - localOffset;
             int hash = data.get(ValueLayout.JAVA_INT_UNALIGNED, currOffset);
             if (semicolonPos < Integer.BYTES) {
                 hash = (byte) hash;
@@ -245,7 +246,6 @@ public class CalculateAverage_merykitty {
                     continue;
                 }
 
-                int localOffset = (int) (currOffset - offset);
                 var nodeKey = ByteVector.fromArray(BYTE_SPECIES, node.data, BYTE_SPECIES.length() - localOffset);
                 var eqMask = line.compare(VectorOperators.EQ, nodeKey).toLong();
                 long validMask = (-1L >>> -semicolonPos) << localOffset;
@@ -256,7 +256,7 @@ public class CalculateAverage_merykitty {
             }
 
             long nextOffset = parseDataPoint(aggr, data, currOffset + 1 + semicolonPos);
-            semicolons >>>= (nextOffset - currOffset);
+            semicolons &= (semicolons - 1);
             if (semicolons == 0) {
                 return nextOffset;
             }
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Oh, wow, this looks very promising. 7.379s on the eval env, fastest so far! Summoning the current top of the pack @royvanrijn, @ebarlas, @spullara, for feedback too!

merykitty:
I just fixed an edge case of overlapping batches, that does not affect the overall performance, though, but you may want to run again just to be sure. Running this on my machine shows that the throughput is bottlenecked by the performance of the `mmap` call. I theorise it may be better to fetch the file concurrently with the processing in a manner similar to prefetching but have not been able to materialise the idea.

image

merykitty:
It turns out that just prefetching the file manually helps lower the time by 20% on my machine, now it takes around 2 seconds, for comparison the baseline takes around 106 seconds and just reading the whole file without doing anything takes around 0.75 seconds.

Thanks.

src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
```diff
@@ -0,0 +1,368 @@
...
+        }
+        point = negative ? -point : point;
+        aggr.min = Math.min(point, aggr.min);
+        aggr.max = Math.max(point, aggr.max);
+        aggr.sum += point;
+        aggr.count++;
+        return offset;
+    }
+
+    private static long iterate(PoorManMap aggrMap, MemorySegment data, long offset) {
```
gunnarmorling, line 207: Could you provide a description of how that iteration loop works? I roughly get it, but it's quite dense code and having a high-level description would be great.

src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
```diff
@@ -0,0 +1,368 @@
...
+        long word = data.get(JAVA_LONG_LT, offset);
+        // This can be 12, 20, 28
+        int decimalSepPos = Long.numberOfTrailingZeros(~word & 0x10101000);
+        int shift = 28 - decimalSepPos;
+        // signed is -1 if negative, 0 otherwise
+        long signed = (~word << 59) >> 63;
+        long designMask = ~(signed & 0xFF);
+        long digits = ((word & designMask) << shift) & 0x0F000F0F00L;
+
+        // Now digits is in the form 0xUU00TTHH00
```
gunnarmorling, line 164: What is TT, HH, etc? As for the main loop, could you provide an overview of how this thing works?

gunnarmorling (commented):
Very nice! This comes in at 7.6, i.e. first place currently! I've added some questions in line to further my (and others) understanding. It looks alright to me, but getting some more intuition would be great. Thanks a lot!

merykitty:
After trying again without prefetching I am also able to reach 2s, which is weird.

Thanks for your reviews, hope I have addressed those.

src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
merykitty: UU stands for the value of the units digit, TT is the tens digit and HH is the hundreds digit.

src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
merykitty: Done

src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
```diff
@@ -0,0 +1,368 @@
...
+        point = negative ? -point : point;
+        aggr.min = Math.min(point, aggr.min);
+        aggr.max = Math.max(point, aggr.max);
+        aggr.sum += point;
+        aggr.count++;
+        return offset;
+    }
+
+    private static long iterate(PoorManMap aggrMap, MemorySegment data, long offset) {
+        var line = ByteVector.fromMemorySegment(BYTE_SPECIES, data, offset, ByteOrder.nativeOrder());
```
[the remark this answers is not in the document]
merykitty: No if we cannot find the delimiter which means the key is longer than the segment we fall back to slow path which do so in a more straightforward way,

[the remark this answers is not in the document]
gunnarmorling: Ah yes, got it. Thanks for adding all the comments!

gunnarmorling:
00:07.620, current #1! Very nice, thanks for participating!

DamienOReilly:
How does this one fare our with Graal?

PR #118, opened 2024-01-05
rby: Has some interesting optimisations but could be improved further with a custom hash map

rby:
- segments the file to number of cores partitions 
- uses ints instead of doubles
- dumb state machine for parsing
- TODO: a custom hash map could improve the updating operation
- ✅ passes ./test.sh rby

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_rby.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_rby.java
@@ -35,6 +35,9 @@ public class CalculateAverage_rby {
         var channel = (FileChannel) Files.newByteChannel(p, EnumSet.of(StandardOpenOption.READ));
         final long size = channel.size();
 
+        if (size < 10000l) {
+            return new long[]{ 0l, size };
+        }
         long chunk = size / workers;
         long position = size - chunk;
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_rby.java
```diff
@@ -0,0 +1,223 @@
...
+                System.out.print(", " + e.getKey() + "="
+                        + (stats[ix]/10.0) + "/"
+                        + avg + "/"
+                        + (stats[ix + 1]/10.0)) ;
+            }
+            System.out.println("}");
+        }
+    }
+
+    static final int MAX_CITIES = 1000;
```
gunnarmorling, line 133: Hum hum, there can be 10,000 station names. How comes that measurements-10000-unique-keys.txt is passing for this one? CC @AlexanderYastrebov

src/main/java/dev/morling/onebrc/CalculateAverage_rby.java
```diff
@@ -0,0 +1,223 @@
...
+                            strbuffIx = 0;
+                            switch (nextChar) {
+                                case ';':
+                                    nextChar = '\n';
+                                    var mbCityIx = cityIndexes.get(str);
+                                    if (mbCityIx == null) {
+                                        cityIndex = nextCityIx;
+                                        cityIndexes.put(str, nextCityIx++);
+                                        if (nextCityIx * 4 >= stats.length) {
+                                            var newStats = Arrays.copyOf(stats, stats.length * 2);
```
rby, line 185: @gunnarmorling This is where more than 1000 cities is handled.
the array is expanded and copied.
In the main case of 1b with real cities names the number was less than 1000

src/main/java/dev/morling/onebrc/CalculateAverage_rby.java
rby: Hi @gunnarmorling, thanks for checking this. In line 185, the array is expanded if there's an overflow.

gunnarmorling: Ah yes, got it. Thanks!

gunnarmorling:
36.518sec. Thx for participating!

PR #119, opened 2024-01-05
More optimizations for ebarlas submission

ebarlas:
Implementation class: `CalculateAverage_ebarlas`

Runtime on `c5.2xlarge` EC2 instance:
```
real	0m7.920s
user	1m0.896s
sys	0m0.664s
```

I added the following line in an attempt to smoothen the sdk selection process. Please be sure GraalVM is selected as it provides a noticeable speed-up over base OpenJDK.

```
```

### DIFF
--- a/calculate_average_ebarlas.sh
+++ b/calculate_average_ebarlas.sh
@@ -15,6 +15,7 @@
 #  limitations under the License.
 #
 
+source "$HOME/.sdkman/bin/sdkman-init.sh"
 sdk use java 21.0.1-graalce
 JAVA_OPTS=""
 time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_ebarlas measurements.txt 8
--- a/calculate_average_ebarlas.sh
+++ b/calculate_average_ebarlas.sh
@@ -16,6 +16,6 @@
 #
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
-sdk use java 21.0.1-graalce
+sdk use java 21.0.1-graalce 1>&2
 JAVA_OPTS=""
 time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_ebarlas measurements.txt 8
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_ebarlas.java
@@ -16,6 +16,7 @@
 package dev.morling.onebrc;
 
 import java.io.IOException;
+import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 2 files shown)


gunnarmorling:
Could you squash all that into a single commit, based off of the latest main? Will make it a bit simpler for me to handle. Thx!

ebarlas:
Sure, I'll do that now.

ebarlas:
Done, @gunnarmorling !

gunnarmorling:
Wow, seeing a massive improvement with this one over your version on main. Do you see that in your env too? What makes the difference?

calculate_average_ebarlas.sh
```diff
@@ -15,6 +15,7 @@
 #  limitations under the License.
 #
 
+source "$HOME/.sdkman/bin/sdkman-init.sh"
 sdk use java 21.0.1-graalce
```
gunnarmorling, line 19:
```suggestion
sdk use java 21.0.1-graalce 1>&2
```

This will make sure that the additional output doesn't interfere with the test assertion.

ebarlas:
Yes! Absolutely massive, as you can see in my timing: 7.x seconds.

I actually provisioned a CX33 VM and it reliably runs at 6.5 seconds there.

ebarlas:
The loop unrolling was the latest leap forward.

ebarlas:
But also, I'm still not quite able to reconcile the _prior_ timing with your reports on my equivalently size EC2 instance. That is, in part, why I provisioned a CX33 VM of my own.

gunnarmorling:
> I actually provisioned a CX33 VM and it reliably runs at 6.5 seconds there.

Oh, wow, that's weird. I see 10-11 sec as of this PR 🤔 . To be sure, I don't think there's CX33, only CCX33. Is that what you have? I definitely see stable results on my machine (just ran yours from main again, it's the same value as currently is on the leaderboard). I could see some variance between machines of the same spec, but 40% seems crazy.

ebarlas:
Darn! I was hoping for sub-10s!

Yes, `CCX33`. I chalked it us to huge variable in the underlying physical machines. My VM is in US. I may try VMs in other geos to try to approach your results.

Here's my last run on CXX33 VM:

```
[root@fedora-32gb-hil-1 ~]# for i in {1..5}; do time java $JAVA_OPTS --class-path average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_ebarlas measurements.txt 8 > out.txt; done

real	0m6.587s
user	0m47.744s
sys	0m1.336s

real	0m6.577s
user	0m47.831s
sys	0m1.278s

real	0m6.550s
user	0m47.730s
sys	0m1.251s

real	0m6.568s
user	0m47.673s
sys	0m1.336s

real	0m6.544s
user	0m47.709s
sys	0m1.320s
```

gunnarmorling:
> Darn! I was hoping for sub-10s!

Pretty sure we'll get there :)

>  I may try VMs in other geos to try to approach your results.

Mine is in nbg1-dc3. They might well have newer boxes in US, as those DCs are newer. It might make sense, once dust has settled a bit, and there's a clear top ten or so, to re-run them again in a fresh box. Let's see. What unnerves me now a little bit, seeing this stark difference, is the risk to lose my current box as it'd mean all existing results are moot. While I have automated the eval pipeline quite a bit by now, I'm not looking forward to doing all that again. Hoping for the best :)

Btw. I have some Terraform set-up [here](https://github.com/gunnarmorling/cloud-boxes/), just in case it's interesting to you.

ebarlas:
Exactly. The instability really came into focus when I got wildly different results on the same VM class. I immediately wondered what _your_ setup is. And also if they can even migrate VMs to different physical hardware.

gunnarmorling:
> And also if Hetzner can migrate VMs to different physical hardware!

At least not that easily I suppose, as there's a local disk attached.

Coming in at 00:10.870. First place! Let's see what @royvanrijn's latest one has to offer :)

PR #128, opened 2024-01-05
Found some extra speed

royvanrijn:
This version is compatible with randomly generated names (other PR) and has some extra tricks to gain a lot of speed, should take over first place.


gunnarmorling:
Could you squash all that into a single commit, based off of the latest main? Will make it a bit simpler for me to handle. Thx!

PR #129, opened 2024-01-05
Entry from gabrielreid

gabrielreid:
Adds a new implementation class `dev.morling.onebrc.CalculateAverage_gabrielreid`, and script for running it with `calculate_average_gabrielreid.sh`.

Somewhat mixed collection of multiple ideas, mostly based initially on using the new JDK Vector API for extracting offsets of newlines and semicolons. Possibly notably, this does not used memory mapped buffers (although I'm doubtful if that would help much here).

Runs locally in just under 11 seconds on 1B rows of input on a 2020 M1 Macbook Air.

### DIFF
--- a/calculate_average_gabrielreid.sh
+++ b/calculate_average_gabrielreid.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:19.561, nice! Thanks for participating!

PR #131, opened 2024-01-05
my small contirbution (in particular the dont dos ;))

rmannibucau:
Since we exchanged on twitter about that here is my version ;)

### DIFF
--- a/calculate_average_rmannibucau.sh
+++ b/calculate_average_rmannibucau.sh
@@ -1,0 +1,53 @@
+#! /bin/bash
+#
+#  Copyright 2024 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# sdk use 21-zulu
+
+# $script build -> build on 21-zulu before running
+# $script native -> use native-image and build before running
+
+function run() {
+  JAVA_OPTS="--enable-preview"
+  java $JAVA_OPTS --class-path target/classes dev.morling.onebrc.CalculateAverage_rmannibucau
+}
+
+if [ -n "$1" ]; then
+  source "$HOME/.sdkman/bin/sdkman-init.sh"
+  case "$1" in
+    build)
+      sdk use java 21-zulu
+      mvn package
+      time run
+      ;;
+
+    # not really faster than jvm mode as of today so no need to test it
+    native)
+      # note: this is not great to native image all mains at once but no need to break the challenge proposal for that yet
+      sdk use java 21.0.1-graal
+      mvn compile
+      native-image --gc=G1 -classpath target/classes dev.morling.onebrc.CalculateAverage_rmannibucau ./target/CalculateAverage_rmannibucau
+      time ./target/CalculateAverage_rmannibucau
+      ;;
+
+    *)
+      echo "[SEVERE] Unknown parameter $1" 1>&2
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Hey @rmannibucau, could you resolve those merge conflicts, rebase to latest main, and squash everything into one commit? Also please make sure that `./test.sh rmannibucau` passes. Thx!

rmannibucau:
hi @gunnarmorling fixed it normally

side note: got some license header issues for some other classes, didn't fix it to limit the diff

gunnarmorling:
> side note: got some license header issues for some other classes, didn't fix it to limit the diff

Interesting. Any actual build failure due to that? Can't seem to reproduce this as of current main.

src/main/java/dev/morling/onebrc/CalculateAverage_rmannibucau.java
```diff
@@ -0,0 +1,736 @@
...
+
+    // NOTE: this is a work code which shouldn't stay but kept to explain the rational behind some impl choice (in Bytes) for the exercise
+    // we assume the station set is closed as of today and optimize Bytes equals impl by checking conflicts only per hash
+    public static class ComputePotentialStationDiffsPerHash {
+        private ComputePotentialStationDiffsPerHash() {
+            // no-op
+        }
+
+        public static void main(final String... args) {
+            final var stations = List.of("Abha",
```
gunnarmorling, line 304: Hum, there should be no assumptions about the specific data set (see rules). Not quite clear to me what's happening here?

calculate_average_rmannibucau.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# $script build -> build on 21-zulu before running
+# $script native -> use native-image and build before running
+
+function run() {
+  JAVA_OPTS="--enable-preview"
+  java $JAVA_OPTS --class-path target/classes dev.morling.onebrc.CalculateAverage_rmannibucau
+}
+
+if [ -n "$1" ]; then
```
gunnarmorling, line 28: So do you want to submit this as native or JVM? If native, please extract the build step into a separate script. See Thomas Würthinger's implementation for an example.

gunnarmorling (changes requested):

rmannibucau:
@gunnarmorling got fixed with some rebasing, guess was due to some diff and fixed since, forget it (the license thing)

src/main/java/dev/morling/onebrc/CalculateAverage_rmannibucau.java
rmannibucau: well, part of what I'm interesting in there is the extreme abuses of challenges on one side but also the fact that some choices don't work better than default impl (linear hash based structures here).
This main (which is totally useless for the runtime, as written, is really to explain why the impl works) is intended to ensure the hashCode distributes the (pseudo-expected) values properly for a hashmap.
Concretely the idea was to validate "offline" Bytes#equals implementation assuming the names are (pseudo)open but compatible with such an implementation - which is generally the case with such input data.

side note: this was also validated with https://github.com/meteostat/weather-stations (21K+ station names) and even if conflicts are possible they are unlikely enough - or would probably be handled in the file generator (depends a bit the ecosystem around) so for me it is not a so unfair assumption.

That said I can totally understand you think it does not respect the rule, no big deal to not integrate the PR for me.
Don't think I would be in the top results anyway in 20 days with this code and just used holidays time to participate ;).

calculate_average_rmannibucau.sh
rmannibucau: jvm, graalvm boost is not impressive enough to be relevant IMHO - it is literally in the error zone - (was some ms on my tests/computer, just kept it in the script to enable it and potentially tune it more but default run is really what should be evaluated)

src/main/java/dev/morling/onebrc/CalculateAverage_rmannibucau.java
gunnarmorling: Yeah, indeed I am inclined to close the PR in its current form, as there shouldn't be any specific asumptions or handling of the example station names from the data set, as per the rules. Thanks a lot for participating nevertheless!

gunnarmorling:
Closing, as discussed.

PR #133, opened 2024-01-05
Fixed graalce typo

crush-157:
Replaced `graalce` with `graal` in leaderboard JDK column for @spullara and @royvanrijn entries as per their `calculate_average_<github_id>.sh` files.

In both cases they run `sdk use java 21.0.1-graal` (i.e. not `graalce`).


gunnarmorling:
Thanks, good catch! In fact, I actually had run @spullara with graalce at first by accident. I have re-run this entry with graal and updated the result accordingly. I already had fixed the wrong SDK when doing the latest run for @royvanrijn. So all looking good now.

PR #140, opened 2024-01-05
Added "magic" implementation

peter-lawrey:
This takes 9.15 seconds with 8 CPU Ryzen 5950X c.f 21s for the previous implementation.


royvanrijn:
Haha... this is against the rules, but interesting nonetheless 😉

PR #141, opened 2024-01-05
Adding more speed improvements, going for first again.

royvanrijn:
Should work with random names as well.

Squashed all commits.

### DIFF
--- a/calculate_average_royvanrijn.sh
+++ b/calculate_average_royvanrijn.sh
@@ -15,7 +15,8 @@
 #  limitations under the License.
 #
 
-sdk use java 21.0.1-graal
+source "$HOME/.sdkman/bin/sdkman-init.sh"
+sdk use java 21.0.1-graal 1>&2
 # Added for fun, doesn't seem to be making a difference...
 if [ -f "target/calculate_average_royvanrijn.jsa" ]; then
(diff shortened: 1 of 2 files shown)


thomaswue:
Did you try native image on your solution as well? I think it could very well help.

```
sdk use java 21.0.1-graal
native-image -O3 -gc=epsilon -march=native -cp target/average-1.0.0-SNAPSHOT.jar -o image_calculateaverage_royvanrijn ev.morling.onebrc.CalculateAverage_royvanrijn
./image_calculateaverage_royvanrijn
```

royvanrijn:
@thomaswue I have tried native compilation, including copying your bash script with pgo, but it wasn't faster than just running the `java` command (on my machine).

gunnarmorling:
@royvanrijn, can you adjust your launch script like this:

```
```

The former makes sure `sdk` is available, the latter prevents the additional output to mess with the test assertion. Thx!

gunnarmorling:
00:07.999, wow, very nice! The implementation LGTM, can't spot anything problematic. @ebarlas, sorry, this was a short stint on #1 :) // CC @spullara

thomaswue:
OK. Adjusted the source script also on my branch and merged with latest master @gunnarmorling.

@royvanrijn I think one main improvement here is to go for the sun.misc.Unsafe usage. Then you should get into the range where it is fast enough for native image to make a difference.

PR #142, opened 2024-01-05
CalculateAverage_truelive second attempt

truelive:
Just removing double parsing shaves off alot.
My implementation: `CalculateAverage_truelive`
Was using `Windows 11` and `GraalVM 21`.

Also GraalVm just magically boosts the whole process. Both GraalVm and OpenJDK spent 100% CPU.

Locally `6.5s` average, AMD Ryzen 9 3900X (12 cores), 64GB RAM. I would expect 20s leaderboard time =)

### DIFF
--- a/calculate_average_truelive.sh
+++ b/calculate_average_truelive.sh
@@ -15,6 +15,8 @@
 #  limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:28.167, thx!

PR #144, opened 2024-01-05
Attempt at the challenge :)

iziamos:
implementation class name: CalculateAverage_iziamos

timing: local time: 37.20user 1.21system 0:01.93elapsed 1986%CPU (0avgtext+0avgdata 3527088maxresident)k

hardware: intel 13700k with 32gb ddr5

Please if you could be kind enough try running with EpsilonGC as well. I haven't defaulted to that because its output appears to affect the test.


iziamos:
oh, no graal because i dont know how to do that

src/main/java/dev/morling/onebrc/CalculateAverage_iziamos.java
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        return result;
+    }
+
+    private static void processEvent(final ByteBuffer buffer, final PartialResult map) {
+        final long hash = parseId(buffer);
+        final int value = readValue(buffer);
+
+        map.recordResult(hash, value);
```
gunnarmorling, line 143: How do you deal with hash collisions here?

gunnarmorling:
One question inline. And could you squash everything into one commit, please? Thx!

src/main/java/dev/morling/onebrc/CalculateAverage_iziamos.java
iziamos: I deliberately ignore that. I'm guessing we need something in the rules maybe. I'll make changes

Issue #35, opened 2024-01-03
Should the 1 billion row file be deterministic?


datdenkikniet:
Currently it seems that the 1 billion rows file is generated randomly. Making the generation pseudorandom would make sharing the 1 billion row file a little easier, and would make sure that everyone is running exactly the same test.

Just using a `Random` with a predefined seed to pick out numbers, and seeding a `Random` with the hash code of the city name to obtain measurements should do the trick.

datdenkikniet:
I've now opened a PR that adds this functionality in #149. Also puts in a little bit of ground work to hopefully make #125 a bit easier to use generically.

mtopolnik:
The evaluation shouldn't use a public test file because that allows the contenders to tightly optimize for the exact keyset in that file. For example, tweaking the hash function to minimize collisions, having special cases for some keys, sizing everything exactly right for the keyset, etc.

mtopolnik:
If there's concern that some solution may just get unlucky with a given keyset, the winner can be determined by repeating the test with 2-3 different test files. I very much doubt that this would be a factor, given the large keyset size (10,000); more noise can be expected from all the environmental factors on the test machine.

datdenkikniet:
I agree! I do not think that the current test-file should be shared or changed, but am asking for determinism so that it becomes a lot easier to compare/run on 1 billion row files that other contestants are using without requiring transmission of the entire data file.

PR #149, opened 2024-01-05
Make generation of measurements file deterministic

datdenkikniet:
Addresses #35.

This is currently not implemented/added for CreateMeasurements2. If it's desirable that both produce the same output, we can add that.


src/main/java/dev/morling/onebrc/WeatherStation.java
```diff
@@ -0,0 +1,518 @@
...
+    /**
+     * We write out data into the buffer to avoid string conversion
+     * We also no longer use double and gaussian, because for our
+     * purpose, the fake numbers here will do it. Less
+     *
+     * @param buffer the buffer to append to
+     */
+    public void measurement(final CheaperCharBuffer buffer) {
+
+        // fake -10.9 to +10.9 variance without double operations and rounding
```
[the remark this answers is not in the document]
datdenkikniet: Ah, okay: using `java.util.Random` instead of `FastRandom` causes this path to be "only" 1.5x faster, as opposed to about 3.5x faster (from my limited measurement)

rschwietzke:
Feel free to undo the use of FastRandom. nextGaussian is a lot of code, hence I prefered to take that all out and also the double to char/byte conversion. There are ways to make that all faster, It is for instance a single thread and also we roll char to byte again and again.

datdenkikniet:
`FastRandom` was neatly quick, and just using it everywhere was quite easy (and provided a bit of speed up to the non-2 version as well). Since both implementations produce the same output now, I vote that we get rid of the slower one.

datdenkikniet:
Updated with the new random station generation.

Now we have 3 measurement creators that output the exact same file :P

I have also made some slight syntactic changes to the name generation @mtopolnik. Semantically it should do the same thing, but would appreciate if you could take a look either way.

image

src/main/java/dev/morling/onebrc/WeatherStationFactory.java
```diff
@@ -0,0 +1,144 @@
...
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+
+import org.rschwietzke.FastRandom;
+
+public class WeatherStationFactory {
```
datdenkikniet, line 30: Generally not too big a fan of `Factory` classes, but this one felt reasonable since it's relatively complicated math stuff.

src/main/java/dev/morling/onebrc/WeatherStationFactory.java
```diff
@@ -0,0 +1,144 @@
...
+                    nameBuf.deleteCharAt(nameBuf.length() - 1);
+                    if (Character.isWhitespace(nameBuf.charAt(nameBuf.length() - 1))) {
+                        nameBuf.setCharAt(nameBuf.length() - 1, readNonSpace(nameSource));
+                    }
+                    name = nameBuf.toString();
+                    actualLen = name.getBytes(StandardCharsets.UTF_8).length;
+                }
+
+                while (names.contains(name)) {
+                    nameBuf.setCharAt(rnd.nextInt(nameBuf.length()), readNonSpace(nameSource));
```
mtopolnik, line 105: Here's the trouble: this change may result in the encoded string exceeding 100 bytes.

datdenkikniet: Ah, right. Characters can be > 1 byte. Will fix

datdenkikniet: Fixed. Now loops `while (names.contains(name) || nameByteLen > 100)` and takes the appropriate action.

src/main/java/dev/morling/onebrc/WeatherStationFactory.java
```diff
@@ -0,0 +1,148 @@
...
+                    nameBuf.setCharAt(0, readNonSpace(nameSource));
+                }
+                if (Character.isWhitespace(nameBuf.charAt(nameBuf.length() - 1))) {
+                    nameBuf.setCharAt(nameBuf.length() - 1, readNonSpace(nameSource));
+                }
+
+                var name = nameBuf.toString();
+                int nameByteLen = name.getBytes(StandardCharsets.UTF_8).length;
+
+                while (names.contains(name) || nameByteLen > 100) {
```
mtopolnik, line 96: Looks good!

src/main/java/dev/morling/onebrc/WeatherStationFactory.java
```diff
@@ -0,0 +1,148 @@
...
+    private static char readNonSpace(StringReader nameSource) throws Exception {
+        char c;
+        do {
+            var n = nameSource.read();
+            if (n == -1) {
+                // TODO: Reset nameSource and read instead of throwing?
+                throw new Exception("Name source exhausted");
+            }
+            c = (char) n;
+        } while (c == ' ');
```
mtopolnik, line 144: As an OCD moment, I'd like to change this to `while (Character.isWhitespace(c))` to align with the other tests.

datdenkikniet: Good catch! Updated

mtopolnik:
I bit the bullet and rented out a CCX3 :D Very surprising results using the full 1B dataset on this machine:

merykitty - 51 seconds
thomaswue - 21 seconds
ebarlas, royvanrijn break on the 31-bit mapped region size length
obourgain: 5 seconds!

mtopolnik:
Correction: didn't check the command output... obourgain takes 5 seconds to break with OOM :D

mtopolnik:
Update: royvanrijn's solution now works.

@gunnarmorling I'm wondering if you're having second thoughts on merging this, given the difference in the timings. OTOH it's kind of beside the point to reward solutions that work great for 416 short keys but degrade significantly for 10,000 keys using the full 100 byte range.

PR #160, opened 2024-01-05
Ensure newline consistency between platforms

### LINKED ISSUE #159: Always use \n as line endings in measurements.txt for uniformity
This line here: https://github.com/gunnarmorling/1brc/blob/c24bcac047d8b7b67115e4420348e4c9410bafdb/src/main/java/dev/morling/onebrc/CreateMeasurements.java#L500

...is platform dependent. When running on windows it generates `\r\n` line endings, which threw me off a bit at first as some of the existing submissions weren't working on my machine.

Could we change it to `bw.write('\n');` so that the line endings are consistent and the same format file is thus generated on all platforms?

(Happy to submit a PR, but it'd obviously be trivial.)

berry120:
Fix #159


gunnarmorling:
Thx!

PR #162, opened 2024-01-06
Improvement in CalculateAverage_yavuztas

yavuztas:
Hi @gunnarmorling 

I improved my solution inspired by the famous @peter-lawrey's solution. It should be 2-3 seconds faster now :)

The changes are in general:
- improve double reading by eliminating string parsing in between
- make calculations over on integer instead of double
- parse into double at the end only once
- remove the character limit per line, it works now independent of the line length
 
Could you please reevaluate when you have time?

Many thanks!

#### Check List:
* Execution time: ~8s (on Macbook Pro m2 16 Core 32 GB Ram)
* Execution time of reference implementation: which one? baseline?? It's 3m 22s


yavuztas:
New execution time after micro optimizations: ~6s
Yay!

gunnarmorling:
00:13.277. Thx for participating!

PR #163, opened 2024-01-06
seijikun: Fix new unit-test introduced with #125

seijikun:
I tried to refactor this code for hours now, but whenever I touch anything in `readStationName()`, the performance deteriorates massively, without any identifiable reason.

Seems like I accidentally hit a magic sweetspot in the C2 Hotspot compiler Vector API support stuff in the way that the code is structured now.


gunnarmorling:
Thx. Results didn't change due to that.

PR #166, opened 2024-01-06
Add solution by flippingbits - Use SIMD for computing aggregates

flippingbits:
First of all, thank you very much for setting up this interesting challenge! I have not used SIMD instructions with Java before, so it provided a nice playground for trying out the `jdk.incubator.vector` module.

The solution approaches the problem as follows:
* Use memory-mapped file to speed up loading data into memory
* Partition data, compute aggregates for partitions in parallel, and finally combine results from all partitions
* Apply SIMD instructions for computing min/max/sum aggregates of partitions
* Use Shorts for storing aggregates, so we maximize the SIMD parallelism

I use the JDK `21.0.1-graal`.

#### Check List:
* Execution time:
```
real 0m9.640s
user 1m11.435s
sys  0m5.286s
```
```
real  3m44.874s
user  3m39.482s
sys   0m6.027s
```

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_flippingbits.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_flippingbits.java
@@ -1,0 +1,211 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_flippingbits.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private static class PartitionAggregate {
+        final short[] lane = new short[SIMD_LANE_LENGTH * 2];
+        // Assume that we do not have more than Integer.MAX_VALUE measurements for the same station per partition
+        int count = 0;
+        long sum = 0;
+        short min = Short.MAX_VALUE;
+        short max = Short.MIN_VALUE;
+
+        public void addMeasurementAndComputeAggregate(byte[] measurementBytes) {
```
gunnarmorling, line 148: Nice, I think it's the first contribution I'm seeing to use vector ops for the book keeping here.

gunnarmorling:
00:26.253. Thx for participating!

PR #167, opened 2024-01-06
armandino: second submission

armandino:
#### Check List:
* Execution time:
```
real    0m44.224s
user    1m14.223s
sys     0m7.502s
```
* Execution time of reference implementation: ~5min

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_armandino.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_armandino.java
@@ -18,164 +18,113 @@ package dev.morling.onebrc;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_armandino.java
```diff
@@ -18,173 +18,111 @@
...
-                }
-                else if (b == MINUS) {
-                    sample.sign = -1;
-                }
-                else {
-                    sample.pushMeasurement(b);
-                }
+            if (stats == null) {
+                stats = new Stats(new String(sample.keyBytes, 0, sample.keyLength, UTF_8));
+                map.put(sample.keyHash, stats);
```
gunnarmorling, line 108: This seems prone to hash collisions? There could be multiple names with the same hash and results would be wrong. I suppose that actually was an issue with the initial implementation already.

gunnarmorling (changes requested):

src/main/java/dev/morling/onebrc/CalculateAverage_armandino.java
armandino: @gunnarmorling thanks for the feedback. This should be equivalent to using `String` as a key if ingot this right. Maybe we need a test for collisions?

gunnarmorling:
Can you please rebase this to current main to resolve the merge conflict? Thx.

armandino:
Having problems with rebase. I'll close this and submit a new PR.

PR #168, opened 2024-01-06
isolgpus: submission 2 - about a 25% improvement on submission 1.

isolgpus:
use as many cores as are available
don't buffer the station name, only use it when we to cache it for the first time. 
get rid of the branching in the main loop
move variables inside the loop

#### Check List:
* Execution time: 1.82 seconds on 5950x
* Execution time of reference implementation: 2 minutes 6 seconds on 5950x


src/main/java/dev/morling/onebrc/CalculateAverage_isolgpus.java
```diff
@@ -123,38 +117,49 @@ private static MeasurementCollector[] handleChunk(FileChannel channel, long esti
...
-                    while ((aChar = r.get()) != SEPERATOR) {
-                        nameBuffer[nameBufferIndex++] = aChar;
-                        nameSum += aChar;
-                        hashResult = 31 * hashResult + aChar;
+                nameStart = i;
+                int nameBufferIndex = 0;
+                int valueIndex = 0;
+
+                // optimistically assume that the name is at least 4 bytes
+                int firstInt = r.getInt();
```
gunnarmorling, line 130: Is this just slow when that assumption breaks, or will it fail?

isolgpus: Just slow, on line 143 if it notices it has gone past the separator without noticing it will rewind back to the beginning of the line and read it byte for byte.

gunnarmorling: Ah yes, I see. Thx!

gunnarmorling:
00:11.222, nice! I suppose the difference on the eval instance is smaller than what you observe as it ran there already with 8 cores before, i.e. this change doesn't make a difference.

PR #175, opened 2024-01-06
Add multithreaded generation of measurements file with Gaussian distribution (M2 AIR took 24 seconds for 1B items)

wlsc:
No changes to the art of generation for the 1st version with nextGaussian function, but now just using all the available threads

### DIFF
--- a/create_measurements_fast.sh
+++ b/create_measurements_fast.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+
+java --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CreateMeasurementsFast $1
--- a/src/main/java/dev/morling/onebrc/CreateMeasurementsFast.java
+++ b/src/main/java/dev/morling/onebrc/CreateMeasurementsFast.java
@@ -1,0 +1,533 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


src/main/java/dev/morling/onebrc/CreateMeasurementsFast.java
```diff
@@ -0,0 +1,533 @@
...
+                new WeatherStation("Yaoundé", 23.8),
+                new WeatherStation("Yellowknife", -4.3),
+                new WeatherStation("Yerevan", 12.4),
+                new WeatherStation("Yinchuan", 9.0),
+                new WeatherStation("Zagreb", 10.7),
+                new WeatherStation("Zanzibar City", 26.0),
+                new WeatherStation("Zürich", 9.3));
+
+        int chunkSize = 10_000_000;
+        int numberOfFutures = size / chunkSize;
```
datdenkikniet, line 506: This rounds down, so trying to produce fewer than 10_000_000 records gives an empty output (because 0 futures are spawned). 

Additionally, when I run this I consistently get fewer lines than requested (e.g. 9_999_999 when I ask for 10_000_000, 99_999_990 when I ask for 100_000_000). Not sure where it comes from, but could you fix that?

create_measurements_fast.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+
+java --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CreateMeasurementsFast $1
```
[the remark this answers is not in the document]
wlsc: yep, added deletion of such file in case exists -- that'd better for convenience reasons ;)

src/main/java/dev/morling/onebrc/CreateMeasurementsFast.java
wlsc: yeah. just now fixed the issue with correct number of lines when generating 1B lines.
btw, why should you generate less than that? challenge is 1B rows ;)

create_measurements_fast.sh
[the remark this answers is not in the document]
datdenkikniet: Sweet, thanks!

src/main/java/dev/morling/onebrc/CreateMeasurementsFast.java
wlsc: anyways, here the fix for less or equal to 10M entries of the chunk size @datdenkikniet

datdenkikniet: Nice :smile:

gunnarmorling:
Nice, thx!

PR #185, opened 2024-01-06
BRC Entry

cliffclick:
#### Check List:
- [X ] Tests pass (`./test.sh <username>`)
- [X ] All formatting changes by the build are committed
- [ X] Output matches that of calculate_average_baseline.sh
* Execution time:
* {2.298,2.438,2.204,2.213,2.823};  Average of middle 3: 2.316s
  2mins, 55sec

My machine: "Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz   3.70 GHz"


gunnarmorling:
Hey @cliffclick, thank you so much for entering this challenge, it's an honor!

Does running `./test.sh cliffclick` complete on your machine? I see it hanging here:

```
Validating calculate_average_cliffclick.sh -- src/test/resources/samples/measurements-10000-unique-keys.txt
```

In addition, two operational requests at this point:

* Could you please make your launch script executable?
* I see you attribute copyright to your entry specifically to you. Is this strictly needed, or can we go with the "original authors" we've used for everything else? I'm fine either way, the license header checker will then just need some configuration to handle this entry specifically.

src/main/java/dev/morling/onebrc/CalculateAverage_cliffclick.java
```diff
@@ -0,0 +1,394 @@
...
+            // F*KING WINDOWS.
+            // Skip CR
+            // idx++;
+
+            w.insert(n8, temp, mmap, cityx);
+        }
+    }
+
+    private static class Work {
+        private static final int TAB_SIZE = 0x800; // 512 for 413 cities
```
bmarwell, line 202: Is this a "legal" assumption according to the rules? 🤔

cliffclick:
Fixed test cases.  Also figured out when the test case checker was failing, because for me on windows it looks like it silently succeeds, when instead it has silently failed.  :-(

My launch script is executable on my platform; I do not know why it would not be for you.

As for the "original authors", I do not know who that is, or what is means.  I wrote *all* the code in the submission, except copying the liscence header and the 1-liner "calculate_average..." script.  I did not copy/clone any code from either the baseline submission, nor any other submission.  So AFAIK, I'm the only author, and I thought that should be explicit.

src/main/java/dev/morling/onebrc/CalculateAverage_cliffclick.java
cliffclick: Yes and no; it fails the 10000 city test case but increasing the table size to 16K fixes that.  Other than that, there's nothing in the rules about supporting more and more cities until memory is exhausted - so a resizing hash table is strictly not required.  If the test cases forced e.g. a million entry table, but the performance work only required a 512 entry table then it might make sense to add the resize logic - as have such a large diffuse table would probably impact caches, and hurt the performance test.

gunnarmorling:
> So AFAIK, I'm the only author, and I thought that should be explicit.

Ok, that is fair. I'll look into configuring the license checker so that this is adhered to. Thanks for clarifying!

src/main/java/dev/morling/onebrc/CalculateAverage_cliffclick.java
gunnarmorling: Yes, this looks ok now. To be valid, it must pass the 10K city test (which it does now), and it must pass evaluation (i.e. yield the expected output) for the original 400 cities key set.

cliffclick:
Fixed bug

gunnarmorling:
All tests passing now and it yields the expected result. 00:04.741 on the evaluation machine, very nice!

PR #201, opened 2024-01-07
More optimizations for ebarlas solution

ebarlas:
#### Check List:
- [x ] Tests pass (`./test.sh ebarlas`)
- [x ] All formatting changes by the build are committed
- [x ] Output matches that of calculate_average_baseline.sh
* Execution time: 6.8s on `c5.2xlarge` (8 vCPU), roughly a 15% reduction in time from prior submission
* Execution time of reference implementation: 3m59s (same VM)

Note that this submission includes a change to use GraalVM rather than GraalVM CE.

### DIFF
--- a/calculate_average_ebarlas.sh
+++ b/calculate_average_ebarlas.sh
@@ -16,6 +16,6 @@
 #
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:09.338, very nice! 3rd place with that right now :)

Issue #101, opened 2024-01-04
Fix tests (ddimtirov)


gunnarmorling:
Hey @ddimtirov, since your submission we've added a basic test suite for covering some corner cases (as we discussed already). One of the tests fails for your submission. Could you take a look and fix this so the submission can be re-evaluated? Thx!

```
./test.sh ddimtirov
Validating calculate_average_ddimtirov.sh -- src/test/resources/samples/measurements-1.txt
PT0.029701S

real	0m0.337s
user	0m0.093s
sys	0m0.130s
1c1
< {=0.0/19.8/19.8, Kunming=0.0/19.8/19.8, g=0.0/19.8/19.8, ing=0.0/19.8/19.8, ming=0.0/19.8/19.8, ng=0.0/19.8/19.8, nming=0.0/19.8/19.8}
---
> {Kunming=19.8/19.8/19.8}
```

ddimtirov:
Will do when I get to updating the submission. It will take a few days.

ddimtirov:
Apologies, the commit was done on Windows and the problem is only visible on Linux/Mac.

Addressing in https://github.com/gunnarmorling/1brc/pull/228

PR #228, opened 2024-01-08
Committing line separator changes on Linux systems 

ddimtirov:
Not applicable - addressing the issue described here: https://github.com/gunnarmorling/1brc/issues/101#issuecomment-1880246100

The root cause is that lacking `.gitattrs` the CMD file was originally committed with Linux file endings. 

Adding `gitattrs` addresed line ending issues manifesting when one shares a FS between Windows and Linux (as is the case with WSL).


gunnarmorling:
Ah yes, good catch. Thx!

PR #245, opened 2024-01-08
Update davecom entry use ByteBuffer instead of String for Names and graal

davecom:
#### Check List:
* Execution time: 17 seconds
* Execution time of reference implementation: 4 minutes


src/main/java/dev/morling/onebrc/CalculateAverage_davecom.java
```diff
@@ -118,14 +120,20 @@ public static void outputResults() {
         // fast string concatenation starting with { and ending with } with a comma between each (no newlines)
         StringBuilder sb = new StringBuilder();
         sb.append('{');
-        var sortedNames = mins.keySet().stream().sorted().toArray(String[]::new);
+        // var sortedNames = mins.keySet().stream().sorted().toArray(String[]::new);
+
         DecimalFormat df = new DecimalFormat("0.0");
```
japplis, line 125: Note that on my French locale, numbers are showing with a comma instead of dot (e.g. -7,5). You may want to use Locale.setDefault() before.
Thanks for your code, I had forgotten about IntSummaryStatistics class. That simplifies my code.

calculate_average_davecom.sh
```diff
@@ -17,4 +17,5 @@
 
 
 JAVA_OPTS=""
+sdk use java 21.0.1-graal 1>&2
```
gunnarmorling, line 20: Could you please move this to a separate file _prepare_davecom.sh_ (see baseline pair of scripts for an example)? It's needed due to a slight reorg to the way we run the entries. Thx!

src/main/java/dev/morling/onebrc/CalculateAverage_davecom.java
davecom: Good point and glad you found it helpful!

calculate_average_davecom.sh
davecom: Sure, I just did that and this pull-request is now updated. Thank you.

gunnarmorling:
00:18.771 now, i.e. no significant difference.

PR #263, opened 2024-01-09
Second tuning for thomaswue

thomaswue:
#### Check List:
* Execution time: 0.80s

This incrementally improves upon the previous version and uses now a long value at a time searching for the delimiter and also for the hash collision check. Improves ~15% over previous version on my machine and runs now < 800ms.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
@@ -31,11 +31,11 @@ import java.util.stream.IntStream;
 /**
  * Simple solution that memory maps the input file, then splits it into one segment per available core and uses
  * sun.misc.Unsafe to directly access the mapped memory. Uses a long at a time when checking for collision.
- *
- * Runs in 0.84s on my Intel i9-13900K
+ * <p>
+ * Runs in 0.80s on my Intel i9-13900K
  * Perf stats:
- *     56,216,315,723      cpu_core/cycles/
- *     67,605,798,701      cpu_atom/cycles/
+ * 53,344,113,388      cpu_core/cycles/
+ * 66,193,198,848      cpu_atom/cycles/
  */
 public class CalculateAverage_thomaswue {
     private static final String FILE = "./measurements.txt";
@@ -81,8 +81,7 @@ public class CalculateAverage_thomaswue {
         // Parallel processing of segments.
         List<HashMap<String, Result>> allResults = IntStream.range(0, chunks.length - 1).mapToObj(chunkIndex -> {
             HashMap<String, Result> cities = HashMap.newHashMap(1 << 10);
-            Result[] results = new Result[1 << 18];
-            parseLoop(chunks[chunkIndex], chunks[chunkIndex + 1], results, cities);
+            parseLoop(chunks[chunkIndex], chunks[chunkIndex + 1], cities);
             return cities;
         }).parallel().toList();
 
@@ -114,7 +113,8 @@ public class CalculateAverage_thomaswue {
         }
     }
 
-    private static void parseLoop(long chunkStart, long chunkEnd, Result[] results, HashMap<String, Result> cities) {
+    private static void parseLoop(long chunkStart, long chunkEnd, HashMap<String, Result> cities) {
+        Result[] results = new Result[1 << 18];
         long scanPtr = chunkStart;
         byte b;
         while (scanPtr < chunkEnd) {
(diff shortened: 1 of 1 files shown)


thomaswue:
@merykitty Your clever number parsing trick is the major aspect missing here I think. Let's join forces ;-).

merykitty:
@thomaswue Yes I think it would be helpful, since the branches are quite unpredictable.

gunnarmorling:
That's awesome, loving this spirit! Exactly what I was hoping to get out of it, folks exchanging with and inspiring each other, which IMO is way more important than who ends up on number #1. Way to go! Still working on rebootstrapping the new eval machine, will get to running this (and all the other pending entries) later this week. Thanks for your patience!

jkroepke:
@gunnarmorling sorry for cross-posting here, but what did you think running an GitHub Runner on the eval machine? and integrate the runner into the CI of the repo?

gunnarmorling:
@jkroepke, I think that's an interesting idea. Could you open an issue for discussing it separately? Thx!

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
```diff
@@ -114,44 +113,44 @@ private static Unsafe initUnsafe() {
...
-                    break;
-                }
-                else if (((nextVal & 0x3B000000) == 0x3B000000)) {
-                    scanPtr += 4;
-                    hash = hash ^ (nextVal & 0xFFFFFF);
-                    break;
+            long hash = 0;
+
+            // Search for ';', one long at a time.
+            long word = UNSAFE.getLong(scanPtr);
```
artsiomkorzun, line 125: This read can go over the end of the file. For example, a file contains the last entry "X;0.0\n". The read goes over 2 bytes at the end of the file. It will touch the next page, if the file has size equal to page size. If the next page is protected, that is seg fault. But I can't reproduce it locally. I always have ~136K memory after the end of the mapped region which I can access witout getting a seg fault. Maybe you have an explaination? 

A workaround is to simply copy the tail with padding (equal has the same flaw and we want it read words).

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
```diff
@@ -175,42 +174,25 @@ else if (((nextVal & 0x3B000000) == 0x3B000000)) {
...
-                        for (; i < (long) nameLength; ++i) {
-                            if (UNSAFE.getByte(existingResult.nameAddress + i) != UNSAFE.getByte(nameAddress + i)) {
-                                result = false;
-                                break;
-                            }
-                        }
-                    }
-                    if (result) {
+                    if (((UNSAFE.getLong(existingResult.nameAddress + i) ^ UNSAFE.getLong(nameAddress + i)) << (64 - (nameLength - i) << 3)) == 0) {
                         existingResult.min = (short) Math.min(existingResult.min, number);
```
BSchneppe, line 196: Looks like these cast would now be unnecessary like in lines 69-70

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
thomaswue: Yes, didn't think about that. It is very unlikely to exactly have a page boundary there, but will have to add the workaround for that last row.

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
thomaswue: Yes, agreed, thanks for spotting. Left over from last version and now this might actually add an additional instruction!

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
```diff
@@ -223,18 +194,47 @@ else if ((long) nameLength >= 4) {
...
+        long value = (absValue ^ signed) - signed;
+        return (int) value;
+    }
+
+    private static int findDelimiter(long word) {
+        long input = word ^ 0x3B3B3B3B3B3B3B3BL;
+        long tmp = (input & 0x7F7F7F7F7F7F7F7FL) + 0x7F7F7F7F7F7F7F7FL;
+        tmp = ~(tmp | input | 0x7F7F7F7F7F7F7F7FL);
+        return Long.numberOfTrailingZeros(tmp) >>> 3;
+    }
```
mukel, line 229: Can be micro-optimized as follows:
```java
private static int findDelimiter(long word) {
    long input = word ^ 0x3B3B3B3B3B3B3B3BL;
    long tmp = (input - 0x0101010101010101L) & ~input & 0x8080808080808080L;
    return Long.numberOfTrailingZeros(tmp) >>> 3;
}
```
Which saves 1 instruction.
Before:                  `139,978,244,139 instructions:u`
With this patch: `134,391,924,137 instructions:u`

merykitty: Since all characters before the delimiter are larger than `0x3B`, you can get away with

    long mask = (word - 0x3C3C3C3C3C3C3C3CL) & 0x8080808080808080L;
    return Long.numberOfTrailingZeros(mask) >>> 3;

Also you need to load this in a little endian manner.

thomaswue: Yes, indeed, thanks. This got another almost 2%.

thomaswue: @gunnarmorling This is already the second contribution from @mukel, I would like to ask if you could add him as a co-author to the submission as well in case you will re-evaluate.

gunnarmorling:
Alrighty, finally back to evaluations :)

This brings your time on the new machine (using 8 of its cores) from 00:03.911 (as per current version on main) down to 00:03.044 as of this branch! Will do a run of the Top 10 on all 64 cores a bit later this months, should be below 1 sec for sure.

I'll add @merykitty and @mukel as co-authors to the entry on the loeaderboard.

thomaswue:
Cool, thank you!

PR #266, opened 2024-01-09
Hyperfine: Script re-org

gunnarmorling:
This splits up the existing launch scripts to adhere to the structure established with https://github.com/gunnarmorling/1brc/pull/182. Only to be merged once that one is in!

@hundredwatt, @AlexanderYastrebov, review welcome. This was a bit of busy work, I hope I didn't mess up anything.


calculate_average_iziamos.sh
```diff
@@ -17,4 +17,4 @@
 
 
 JAVA_OPTS="--enable-preview"
-time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_iziamos
+tjava $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_iziamos
```
AlexanderYastrebov, line 20: tjava

AlexanderYastrebov:
Since you change all of them, maybe it would be an improvement to move them into `./benchmark` or `./bench` before its too late (I think it would be easy to adjust eval scripts).
But this could also be done in subsequent PRs in bulk after this prepare/calculate split.

AlexanderYastrebov:
I think you'd also like to update README.md

gunnarmorling:
Fixed the typo and updated the README. I'm leaving that re-org for a separate PR, to keep things manageable. Thanks a lot for reviewing, @AlexanderYastrebov.

gunnarmorling:
Tested with a few entries, all looking good. Merging.

PR #269, opened 2024-01-09
1brc submission - thanhtrinity

thanhtrinity:
#### Check List:
- [x] Tests pass (`./test.sh thanhtrinity` shows no differences between expected and actual outputs)
- [x] Your launch script is named `calculate_average_thanhtrinity.sh` (make sure to match casing of your GH user name) and is executable
* Execution time:  ~11s on `c5.2xlarge`

When using Memory-Mapped File and FileChanel, ensure that each chunk ends with a newline character before initiating a thread. This approach enhances safety and prevents the merging of multiple headers and footers in each chunk.
I have tried with virtual threads but noticed they were slower compared to standard threads :).

### DIFF
--- a/calculate_average_thanhtrinity.sh
+++ b/calculate_average_thanhtrinity.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+JAVA_OPTS="--enable-preview"
+java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_thanhtrinity
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_thanhtrinity.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_thanhtrinity.java
@@ -1,0 +1,256 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


calculate_average_thanhtrinity.sh
```diff
@@ -0,0 +1,20 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+JAVA_OPTS="--enable-preview"
+sdk use java 21.0.1-graal 1>&2
+time java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_thanhtrinity
```
gunnarmorling, line 20:
```suggestion
java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_thanhtrinity
```

`time` isn't needed any more, and the sdk call should now go into a `prepare_xyz.sh` script (see baseline for an example).

gunnarmorling (changes requested):
Please rebase and apply the suggested changes. Thx!

thanhtrinity:
@gunnarmorling  changed as your suggest, thanks.

calculate_average_thanhtrinity.sh
thanhtrinity: Agreed

gunnarmorling:
This produces a different output than `./calculate_average_baseline.sh`. Can you please investigate?

src/main/java/dev/morling/onebrc/CalculateAverage_thanhtrinity.java
```diff
@@ -0,0 +1,256 @@
...
+        boolean isFractional = false;
+        double divisorForFraction = 1;
+        boolean isNegative = false;
+        for (int i = 0; i < buffer.limit(); i++) {
+            var b = buffer.get();
+            var position = i + 1;
+            if (isProcessKey) {
+                if (b == ';') {
+                    int cIdx = abs(hashKey % citySize);
+                    city = cities[cIdx];
```
gunnarmorling, line 122: How do you deal with hash collisions here? Can't seem to find any sort of linear probing or the like in that case.

thanhtrinity:
> Owner
Thanks for update
I have update code that remove abs and use 0x7FFFFFFF to  avoids any negative and duplicate hash values that could cause issues with array indexing.

gunnarmorling:
Hum, hum. It should not be a question of probability though. You'd have to implement some means of [linear probing](https://en.wikipedia.org/wiki/Linear_probing) to make sure that each aggregation has exactly its own entry in the array. Maybe I'm just missing where this is done?

thanhtrinity:
You're right, the output data 'might' be the same as the baseline output, but there's always a probability of collisions in hash functions, a detail I neglected. I just added linear probing to handle the collisions. Thanks

gunnarmorling:
00:13.013. Thx for participating!

PR #274, opened 2024-01-09
evaluate2.sh improvements - leaderboard, default SDK

### LINKED ISSUE #270: Sort leaderboard by time
When running _./evaluate2.sh_ for multiple entries, the emitted leaderboard should be ordered by decreasing time.

hundredwatt:
Closes #270 

- [x] Sort leaderboard by time
- [x] Show name of Github user (maybe with a local cache 😉)
- [x] we could just fake it by saying "GraalVM native binary" when the additional build script
- [x]  reset the JDK to the default (21.0.1-open) when no prepare script is provided

Example leaderboard output (used 100m rows instead of 1b to save time, that's why everyone looks so fast here 😉)

image

### DIFF
--- a/evaluate2.sh
+++ b/evaluate2.sh
@@ -34,6 +34,8 @@ RED='\033[0;31m'
 BOLD_YELLOW='\033[1;33m'
 RESET='\033[0m' # No Color
 
+DEFAULT_JAVA_VERSION="21.0.1-open"
+
 function check_command_installed {
   if ! [ -x "$(command -v $1)" ]; then
     echo "Error: $1 is not installed." >&2
@@ -56,9 +58,9 @@ fi
 source "$HOME/.sdkman/bin/sdkman-init.sh"
 
 # 3. make sure the default java version is installed
-if [ ! -d "$HOME/.sdkman/candidates/java/21.0.1-open" ]; then
-  echo "+ sdk install java 21.0.1-open"
-  sdk install java 21.0.1-open
+if [ ! -d "$HOME/.sdkman/candidates/java/$DEFAULT_JAVA_VERSION" ]; then
+  echo "+ sdk install java $DEFAULT_JAVA_VERSION"
+  sdk install java $DEFAULT_JAVA_VERSION
 fi
 
 # 4. Install missing SDK java versions in any of the prepare_*.sh scripts for the provided forks
@@ -118,8 +120,8 @@ for fork in "$@"; do
     echo "+ source ./prepare_$fork.sh"
     source "./prepare_$fork.sh"
   else
-    echo "+ sdk use java 21.0.1-open"
-    sdk use java 21.0.1-open
+    echo "+ sdk use java $DEFAULT_JAVA_VERSION"
+    sdk use java $DEFAULT_JAVA_VERSION
   fi
 
   # Optional additional build steps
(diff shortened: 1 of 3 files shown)


evaluate2.sh
```diff
@@ -175,12 +174,50 @@ for fork in "$@"; do
...
-  # var result = String.format("%02d:%02d.%.0f", mean.toMinutesPart(), mean.toSecondsPart(), (double) mean.toNanosPart() / 1_000_000);
-  # var author = actualFile.replace(".out", "")
-  # System.out.println(String.format("\n|   |        %s| [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_%s.java)| 21.0.1-open | [%s](https://github.com/%s)|", result, author, author, author));
+  # Get Github user's name from public Github API (rate limited after ~50 calls, so results are cached in github_users.txt)
+  set +e
+  github_user__name=$(grep "^$fork;" github_users.txt | cut -d ';' -f2)
+  if [ -z "$github_user__name" ]; then
+    github_user__name=$(curl -s https://api.github.com/users/$fork | jq -r '.name' | tr -d '"')
+    if [ "$github_user__name" != "null" ]; then
+      echo "$fork;$github_user__name" >> github_users.txt
```
hundredwatt, line 183: I didn't look into the ToS to see if caching Github data is allowable. Seems innocent, so ask forgiveness not permission 🤷‍♂️

evaluate2.sh
```diff
@@ -175,12 +174,50 @@ for fork in "$@"; do
...
+  # Read java version from prepare_$fork.sh if it exists, otherwise assume 21.0.1-open
+  java_version="21.0.1-open"
+  if [ -f "./prepare_$fork.sh" ]; then
+    java_version=$(grep "sdk use java" ./prepare_$fork.sh | cut -d' ' -f4)
+  fi
+
+  # Hard-coding the note message for now
+  notes=""
+  if [ -f "./additional_build_steps_$fork.sh" ]; then
+    notes="GraalVM native binary"
```
hundredwatt, line 199: Hard-coded notes will need to be changed later if other techniques are used

hundredwatt (commented):
A couple callouts

evaluate2.sh
gunnarmorling: Yeah, that should be fine.

gunnarmorling:
Oh, wow, that was fast, amazing! Loving your approach to writing shell scripts btw, I think it's one of the few instances where I think they are actually legible ;)

We can merge as-is. But just in case, one more thing came to my mind (I know, I know, I'm greedy ;) Could we run `sdk install` for all the SDKs detected in any of the evaluated prepare scripts? AFAIR, the call is a no-op if an SDK is already installed, so it shouldn't add too much overhead while making the entire script pretty much executable as-is, without requiring any more out-of bands preparations (apart from providing the measurements file and expected output for it).

hundredwatt:
> Oh, wow, that was fast, amazing! Loving your approach to writing shell scripts btw, I think it's one of the few instances where I think they are actually legible ;)

Haha, thanks! 😊

> Could we run sdk install once at the beginning of the script for all the SDKs detected in any of the evaluated prepare scripts?

Done! 4ecd342

PR #278, opened 2024-01-10
Fix test.sh to use prepare script

AlexanderYastrebov:
Followup on #266


AlexanderYastrebov:
@gunnarmorling

gunnarmorling:
Ah yes, good catch. I was (am) so busy with the move of the eval script to hyperfine, that I didn't focus on the tests a lot. Thx!

Issue #283, opened 2024-01-10
Validate entry names eagerly in evaluate2.sh


gunnarmorling:
It should be detected at the beginning of the script, whether all passed entry names are valid, i.e. whether a _calculate\_average\_\<entry\>.sh_ exists. And if not, abort the script before running anything.

gunnarmorling:
CC @hundredwatt

gunnarmorling:
To piggyback on this one, the sdk install step should be robust against an empty prepare file. I think right now the grep call fails and aborts the script when it doesn't match anything (can be reproduced by running _evaluate2.sh baseline_.

hundredwatt:
Confirmed, if `prepare_$fork.sh` is present, but has no `sdk use ...`, then `read -r version` fails and exits:

```
+ for fork in '"$@"'
+ '[' -f ./prepare_baseline.sh ']'
+ grep -h '^sdk use' ./prepare_baseline.sh
+ cut '-d ' -f4
+ read -r version
```

PR #289, opened 2024-01-10
evaluate2.sh: validate fork names and fix grep bug

hundredwatt:
Call #283

- [x] Validate fork names
- [x] In block that installs SDKs, grep returns exit code 1 when no match. Add `|| true` to prevent the script from exiting early.

image

### DIFF
--- a/evaluate2.sh
+++ b/evaluate2.sh
@@ -82,7 +82,6 @@ for fork in "$@"; do
     done || true # grep returns exit code 1 when no match, `|| true` prevents the script from exiting early
   fi
 done
-echo "SUCCESS"
 ## END - SDKMAN Setup
 
 # Check if SMT is enabled (we want it disabled)
--- a/evaluate2.sh
+++ b/evaluate2.sh
@@ -47,6 +47,14 @@ check_command_installed java
 check_command_installed hyperfine
 check_command_installed jq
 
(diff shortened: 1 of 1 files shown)


hundredwatt:
cc @gunnarmorling

gunnarmorling:
Excellent, thx!

hundredwatt:
@gunnarmorling FYI this runs all forks:

```sh
./evaluate2.sh `ls -1 ./calculate_average_*.sh | cut -d_ -f3 | cut -d. -f1`
```

I tested this (using a tiny measurements.txt script + adding `timeout 1` before `./calculate_average_.sh`) and found another bug: 1730ad1

So, the script works on my machine at least for all forks! If you want to try running all forks on the eval machine, I recommend adding `timeout` like this to catch forks that hang forever:
```bash
    numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
```

to 

```bash
    timeout 600 numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
```

When a command fails (via timeout or other failure caught by hyperfine), this message appears and the $fork is excluded from the Summary/Leaderboard same as if the output did not match:

image

PR #293, opened 2024-01-10
evaluate2.sh: Add Time Limit for Runs

hundredwatt:
Followup to #290

This is the approach recommended by hyperfine's maintainer: https://github.com/sharkdp/hyperfine/issues/576

### DIFF
--- a/evaluate2.sh
+++ b/evaluate2.sh
@@ -47,6 +47,7 @@ function check_command_installed {
 check_command_installed java
 check_command_installed hyperfine
 check_command_installed jq
+check_command_installed bc
 
 # Validate that ./calculate_average_<fork>.sh exists for each fork
 for fork in "$@"; do
--- a/evaluate2.sh
+++ b/evaluate2.sh
@@ -35,6 +35,7 @@ BOLD_YELLOW='\033[1;33m'
 RESET='\033[0m' # No Color
 
 DEFAULT_JAVA_VERSION="21.0.1-open"
+RUN_TIME_LIMIT=300 # seconds
 
 function check_command_installed {
   if ! [ -x "$(command -v $1)" ]; then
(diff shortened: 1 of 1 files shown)


evaluate2.sh
```diff
@@ -149,9 +150,13 @@ for fork in "$@"; do
 
     # Linux platform
     # prepend this with numactl --physcpubind=0-7 for running it only with 8 cores
-    numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
-  else
-    hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
+    numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "timeout -v $RUN_TIME_LIMIT ./calculate_average_$fork.sh 2>&1"
+  else # MacOS
```
gunnarmorling, line 154: Ah, very nice. So `evaluate2.sh` runs on MacOS too?

hundredwatt: Yes, I've been testing on both Fedora Linux and my MacBook M2 Air

hundredwatt:
Since this has not been merged yet, I added a check that `bc` is installed (it was missing on a test VM I used)

evaluate2.sh
AlexanderYastrebov: Do we need to check whether `gtimeout` command is available?

gunnarmorling:
> Since this has not been merged yet, I added a check that bc is installed (it was missing on a test VM I used)

Nice; just ran into that too.

evaluate2.sh
gunnarmorling: I have `timeout` rather than `gtimeout` on MacOS. Should we just use that?

gunnarmorling: LOL, nevermind, wrong console ;)

It is checked already, so LGTM. Will merge. Will only add log message for this case.

evaluate2.sh
```diff
@@ -149,9 +151,13 @@ for fork in "$@"; do
 
     # Linux platform
     # prepend this with numactl --physcpubind=0-7 for running it only with 8 cores
-    numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
-  else
-    hyperfine $HYPERFINE_OPTS "./calculate_average_$fork.sh 2>&1"
+    numactl --physcpubind=0-7 hyperfine $HYPERFINE_OPTS "timeout -v $RUN_TIME_LIMIT ./calculate_average_$fork.sh 2>&1"
```
AlexanderYastrebov, line 154: I get 
```
timeout: invalid option -- 'v'
Try 'timeout --help' for more information.
```
It works without -v

AlexanderYastrebov: I guess I have outdated version, I see -v here https://man7.org/linux/man-pages/man1/timeout.1.html

PR #301, opened 2024-01-10
Remove additional_build_steps_*.sh support

AlexanderYastrebov:
There is no need to have it as preparation steps could be fit into prepare_*.sh


AlexanderYastrebov:
@hundredwatt @gunnarmorling @thomaswue

evaluate2.sh
```diff
@@ -236,14 +229,14 @@ for fork in "$@"; do
...
   # Hard-coding the note message for now
   notes=""
-  if [ -f "./additional_build_steps_$fork.sh" ]; then
-    notes="GraalVM native binary"
+  if [ -f "./prepare_$fork.sh" ]; then
+    java_version=$(grep -F "sdk use java" ./prepare_$fork.sh | cut -d' ' -f4)
+
+    if grep -F "native-image" -q ./prepare_$fork.sh ; then
+      notes="GraalVM native binary"
+    fi
```
hundredwatt, line 239: Looks good to me!

prepare_thomaswue.sh
```diff
@@ -17,3 +17,9 @@
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
 sdk use java 21.0.1-graal 1>&2
+
+# ./mvnw clean verify removes target/ and will re-trigger native image creation.
+if [ ! -f target/CalculateAverage_thomaswue_image ]; then
```
hundredwatt, line 22: Not sure we want this, what if there was a code change since the last run?

hundredwatt (commented):
Change looks good to me, I don't see any downside

prepare_thomaswue.sh
AlexanderYastrebov: `# ./mvnw clean verify removes target/ and will re-trigger native image creation.`

hundredwatt: Oh, didn't notice you moved the output to the target directory

gunnarmorling:
Yeah, nice improvement. Great to see how the pieces come together. Thx @AlexanderYastrebov, and thx @hundredwatt for reviewing! Merging.

PR #315, opened 2024-01-11
Adding delayed string creation again for (no-) gc, small tweaks

royvanrijn:
#### Check List:
* Execution time: `0m1.333s`
* Execution time of previous implementation: `0m1.370s`

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
@@ -51,7 +51,8 @@ import sun.misc.Unsafe;
  * Fixed bug, UNSAFE bytes String:   1850 ms
  * Separate hash from entries:       1550 ms
  * Various tweaks for Linux/cache    1550 ms (should/could make a difference on target machine)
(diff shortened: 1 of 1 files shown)


royvanrijn:
That’s weird, I’m running mine on an M2 Pro, and it runs in about ~1.35s (native-image).

It’s getting dependent on machine specifics now haha.

ricardopieper:
Yes, very very specifics... 
I tried yours using native image and it's a bit better, 24 seconds. instead of 27.
Mine got twice as slow though, from 10 to 19 seconds.

Perhaps I should run on linux and see how it performs.

ricardopieper:
I just spin up an AWS instance to see how it goes.
c6a-4xlarge, 60GB SSD with provisioned iops (io2).

The results are in line with my Windows PC: my code runs in 11 seconds, yours runs in 6 seconds. I did run a slightly older version of your code though, but still.... interesting results.

gunnarmorling:
00:02.904 on the eval machine, trimmed mean after ten runs. As mentioned before, it's really close to @thomaswue's entry, and I hope we can measure the Top 5 or so on a highly tuned machine (CC @franz1981) towards the end of the challenge, but for now it seems safe enough for our purposes here to put it with the result above to the scoreboard.

PR #317, opened 2024-01-11
Update README - newline not allowed in station names

parkertimmins:
Per https://github.com/gunnarmorling/1brc/discussions/311 station names will not contain newline characters.


gunnarmorling:
Noice, thx!

PR #331, opened 2024-01-11
merykitty gave in to Unsafe

merykitty:
#### Check List:
* Execution time of `merykittyunsafe`:

```
 Performance counter stats for 'sh calculate_average_merykittyunsafe.sh':

          13497.35 msec task-clock:u              #   10.750 CPUs utilized
                 0      context-switches:u        #    0.000 /sec
                 0      cpu-migrations:u          #    0.000 /sec
            227100      page-faults:u             #   16.826 K/sec
       60988772290      cycles:u                  #    4.519 GHz
         289560740      stalled-cycles-frontend:u #    0.47% frontend cycles idle
         114022556      stalled-cycles-backend:u  #    0.19% backend cycles idle
      114074745602      instructions:u            #    1.87  insn per cycle
                                                  #    0.00  stalled cycles per insn
       12762459789      branches:u                #  945.553 M/sec
          53292216      branch-misses:u           #    0.42% of all branches

       1.255587212 seconds time elapsed

      12.568676000 seconds user
       0.870567000 seconds sys
```
 
* Execution time of `merykitty`:

```
 Performance counter stats for 'sh calculate_average_merykitty.sh':

          15482.62 msec task-clock:u              #   11.502 CPUs utilized
                 0      context-switches:u        #    0.000 /sec
                 0      cpu-migrations:u          #    0.000 /sec
            222547      page-faults:u             #   14.374 K/sec
       71001812851      cycles:u                  #    4.586 GHz
         362493890      stalled-cycles-frontend:u #    0.51% frontend cycles idle
         130886807      stalled-cycles-backend:u  #    0.18% backend cycles idle
      143777012786      instructions:u            #    2.02  insn per cycle
                                                  #    0.00  stalled cycles per insn
       23320869203      branches:u                #    1.506 G/sec
          45042342      branch-misses:u           #    0.19% of all branches

       1.346125882 seconds time elapsed

      14.487117000 seconds user
       0.916624000 seconds sys
```

Not sure if this is allowed, I have consciously avoided using `Unsafe`, and this is my attempt to truly push the challenge to its limit.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
@@ -1,0 +1,379 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package dev.morling.onebrc;
+
+import jdk.incubator.vector.ByteVector;
+import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorSpecies;
+import sun.misc.Unsafe;
+
+import java.io.IOException;
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.lang.reflect.Field;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileChannel.MapMode;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class CalculateAverage_merykittyunsafe {
+    private static final String FILE = "./measurements.txt";
+    private static final Unsafe UNSAFE;
+    static {
+        try {
+            Field theUnsafe = Unsafe.class.getDeclaredField("theUnsafe");
+            theUnsafe.setAccessible(true);
+            UNSAFE = (Unsafe) theUnsafe.get(Unsafe.class);
+        }
+        catch (NoSuchFieldException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static final VectorSpecies<Byte> BYTE_SPECIES = ByteVector.SPECIES_PREFERRED.length() >= 32
+            ? ByteVector.SPECIES_256
+            : ByteVector.SPECIES_128;
+    private static final long KEY_MAX_SIZE = 100;
+
+    private static class Aggregator {
+        private long min = Integer.MAX_VALUE;
+        private long max = Integer.MIN_VALUE;
+        private long sum;
+        private long count;
+
+        public String toString() {
+            return round(min / 10.) + "/" + round(sum / (double) (10 * count)) + "/" + round(max / 10.);
+        }
+
+        private double round(double value) {
+            return Math.round(value * 10.0) / 10.0;
+        }
+    }
+
+    // An open-address map that is specialized for this task
+    private static class PoorManMap {
+        // 100-byte key + 4-byte hash + 4-byte size +
+        // 2-byte min + 2-byte max + 8-byte sum + 8-byte count
+        private static final int ENTRY_SIZE = 128;
+        private static final int SIZE_OFFSET = 0;
+        private static final int MIN_OFFSET = 4;
+        private static final int MAX_OFFSET = 6;
+        private static final int SUM_OFFSET = 8;
+        private static final int COUNT_OFFSET = 16;
+        private static final int KEY_OFFSET = 24;
+
+        // There is an assumption that map size <= 10000;
+        private static final int CAPACITY = 1 << 17;
+        private static final int ENTRY_MASK = ENTRY_SIZE * CAPACITY - 1;
+
+        final byte[] data;
+
+        PoorManMap() {
+            this.data = new byte[CAPACITY * ENTRY_SIZE];
+        }
+
+        void observe(long entryOffset, long value) {
+            long baseOffset = Unsafe.ARRAY_BYTE_BASE_OFFSET + entryOffset;
+            UNSAFE.putShort(this.data, baseOffset + MIN_OFFSET,
+                    (short) Math.min(value, UNSAFE.getShort(this.data, baseOffset + MIN_OFFSET)));
+            UNSAFE.putShort(this.data, baseOffset + MAX_OFFSET,
+                    (short) Math.max(value, UNSAFE.getShort(this.data, baseOffset + MAX_OFFSET)));
+            UNSAFE.putLong(this.data, baseOffset + SUM_OFFSET,
+                    value + UNSAFE.getLong(this.data, baseOffset + SUM_OFFSET));
+            UNSAFE.putLong(this.data, baseOffset + COUNT_OFFSET,
+                    1 + UNSAFE.getLong(this.data, baseOffset + COUNT_OFFSET));
+        }
+
+        long indexSimple(long address, int size) {
+            int x;
+            int y;
+            if (size >= Integer.BYTES) {
+                x = UNSAFE.getInt(address);
+                y = UNSAFE.getInt(address + size - Integer.BYTES);
+            }
+            else {
+                x = UNSAFE.getByte(address);
+                y = UNSAFE.getByte(address + size - Byte.BYTES);
+            }
+            int hash = hash(x, y);
+            long entryOffset = (hash * ENTRY_SIZE) & ENTRY_MASK;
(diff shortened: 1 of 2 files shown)


merykitty:
The code consistently outpaces the solutions of Thomas and Roy in C2, but I don't think it will do so after taking into consideration AOT vs JIT and after staring at the assembly for 3 hours I have not come up with any ideas.

artsiomkorzun:
@merykitty How much slower does it perform with aot?

src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
```diff
@@ -0,0 +1,379 @@
...
+        private static final int MAX_OFFSET = 6;
+        private static final int SUM_OFFSET = 8;
+        private static final int COUNT_OFFSET = 16;
+        private static final int KEY_OFFSET = 24;
+
+        // There is an assumption that map size <= 10000;
+        private static final int CAPACITY = 1 << 17;
+        private static final int ENTRY_MASK = ENTRY_SIZE * CAPACITY - 1;
+
+        final byte[] data;
```
franz1981, line 86: Hihi it reminds me of https://github.com/gunnarmorling/1brc/pull/242#issuecomment-1884937902 go go go!

merykitty:
@artsiomkorzun It is about 8 times slower with Graal, I do not have a Graal built with hsdis at hand but looking at the code comments:

      0x00007fadf73244c8: ; ImmutableOopMap {rax=Oop r10=Oop r11=Oop [16]=Oop [24]=Oop }
                          ;*iinc {reexecute=1 rethrow=0 return_oop=0}
                          ; - (reexecute) jdk.incubator.vector.ByteVector::ldLongOp@35 (line 368)
                          ; - jdk.incubator.vector.ByteVector$ByteSpecies::ldLongOp@8 (line 4262)
                          ; - jdk.incubator.vector.ByteVector::lambda$fromMemorySegment0Template$105@8 (line 3811)
                          ; - jdk.incubator.vector.ByteVector$$Lambda/0x00007fad9c05e900::load@10
                          ; - jdk.internal.vm.vector.VectorSupport::load@32 (line 428)
                          ; - jdk.internal.misc.ScopedMemoryAccess::loadFromMemorySegmentScopedInternal@28 (line 361)
                          ; - jdk.internal.misc.ScopedMemoryAccess::loadFromMemorySegment@31 (line 338)
                          ; - jdk.incubator.vector.ByteVector::fromMemorySegment0Template@33 (line 3807)
                          ; - jdk.incubator.vector.Byte256Vector::fromMemorySegment0@3 (line 938)
                          ; - jdk.incubator.vector.ByteVector::fromMemorySegment@31 (line 3297)
                          ; - dev.morling.onebrc.CalculateAverage_merykittyunsafe::iterate@37 (line 256)

I assume that `jdk.internal.vm.vector.VectorSupport::load` is not intrinsified properly, some more other vector intrinsics seem to express the same behaviour.

thomaswue:
@merykitty Yes, we are aware of this issue with the Vector API. Unfortunately it still changes quite often when in incubator phase and it is difficult for us to keep up with the intrinsifications, which for this kind of low-level compiler API are essential.

gunnarmorling:
It doesn't seems to be a clear improvement on the eval machine for some reason. Here's the results from 10 runs:

```
Benchmark 1: timeout -v 300 ./calculate_average_merykitty.sh 2>&1
  Time (mean ± σ):      3.268 s ±  0.127 s    [User: 21.568 s, System: 0.741 s]
  Range (min … max):    2.953 s …  3.430 s    10 runs

Summary
  merykitty: trimmed mean 3.286730414775, raw times 3.1829055004,2.9527902024,3.3237153764,3.3429483424,3.2571899574,3.4296059994,3.2995724264,3.2947934074000003,3.3001821494000003,3.2925361584000004

Leaderboard

|   | 00:03.286 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java)| 21.0.1-open | [Quan Anh Mai](https://github.com/merykitty) |  |

```

merykitty:
@gunnarmorling It is a separate entry `merykittyunsafe` instead of `merykitty`. I don't really want to add `Unsafe` to my original submission so can this go as a separate entry? Thanks.

gunnarmorling:
> @gunnarmorling It is a separate entry merykittyunsafe instead of merykitty.

Ah, sorry, had missed that. Oh, boy 🤯 :

```
Benchmark 1: timeout -v 300 ./calculate_average_merykittyunsafe.sh 2>&1
  Time (mean ± σ):      2.573 s ±  0.024 s    [User: 15.982 s, System: 0.749 s]
  Range (min … max):    2.521 s …  2.605 s    10 runs

Summary
  merykittyunsafe: trimmed mean 2.575439155045, raw times 2.57165876642,2.57405251042,2.55458195942,2.5576553414200003,2.52133204842,2.5802883164200003,2.59440042542,2.58838683042,2.60452826242,2.58248909042

Leaderboard

|   | 00:02.575 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java)| 21.0.1-open | [merykittyunsafe](https://github.com/merykittyunsafe) |  |

```

> I don't really want to add Unsafe to my original submission so can this go as a separate entry? Thanks.

Yes, we can do that. While there should be only one entry per participant by default, I am happy to make an exception for this case.

merykitty:
@gunnarmorling Wow that is really impressive, I guess the test machine is much more capable at utilising the reduction in instruction count than mine. Thanks a lot for your help.

src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
```diff
@@ -0,0 +1,379 @@
...
+                }
+            }
+        }
+
+        void insertInto(long entryOffset, long address, int size) {
+            long baseOffset = Unsafe.ARRAY_BYTE_BASE_OFFSET + entryOffset;
+            UNSAFE.putInt(this.data, baseOffset + SIZE_OFFSET, size);
+            UNSAFE.putShort(this.data, baseOffset + MIN_OFFSET, Short.MAX_VALUE);
+            UNSAFE.putShort(this.data, baseOffset + MAX_OFFSET, Short.MIN_VALUE);
+            try (var arena = Arena.ofConfined()) {
```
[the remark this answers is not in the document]
merykitty: This trick presented by Maurizio Cimadamore depends on the compiler's ability to scalar replace the arena and elide the lifetime check, it will fail miserably if the arena creation is not inlined, though.

thomaswue:
@merykitty Really cool vectorized solution!

On my machine utilizing all CPU cores the solution becomes memory-bound, but I think given that the evaluation is done only on 8 cores out of the 64, the reduction in instructions is giving the speed-up.

mtopolnik:
Strangely enough, both merykitty and merykittyunsafe are very slow on the old Hetzner CCX33. 45 and 40 seconds, respectively (on the official dataset). On the same instance, royvanrijn is at 4.7 seconds. I know the new instance doesn't have AVX-512 support, either, so I wonder what should explain this.

franz1981:
@mtopolnik that's why, in some extent I personally prefer SWAR, when applicable, cause it works the same across archs, although it add some register pressure. Peak performance is obvious not comparable...

merykitty:
@mtopolnik Are you running with C2 or with Graal? Because the latter has not caught up with the development of the Vector API yet.

thomaswue:
Yes, @mtopolnik the numbers there seem to indicate you were running with Graal JIT with the missing intrinsifications for the incubator Vector API.

thomaswue:
Didn't btw know about "-Djdk.incubator.vector.VECTOR_ACCESS_OOB_CHECK=0" yet. Is this planned to stay and be the new unsafe ;-) ?

src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
```diff
@@ -0,0 +1,379 @@
...
+                var arena = Arena.ofShared()) {
+            var data = file.map(MapMode.READ_ONLY, 0, file.size(), arena);
+            long chunkSize = Math.ceilDiv(data.byteSize(), processorCnt);
+            var threadList = new Thread[processorCnt];
+            var resultList = new PoorManMap[processorCnt];
+            for (int i = 0; i < processorCnt; i++) {
+                int index = i;
+                long offset = i * chunkSize;
+                long limit = Math.min((i + 1) * chunkSize, data.byteSize());
+                var thread = new Thread(() -> resultList[index] = processFile(data, offset, limit));
```
yawkat, line 363: Noob here: I've heard in the past that lambda initialization cost is significant compared to anonymous classes. Is that not an issue here?

mtopolnik:
> @mtopolnik Are you running with C2 or with Graal? Because the latter has not caught up with the development of the Vector API yet.

Oops, I have a script that calls `prepare_author.sh` and then `calculate_average_author.sh`. But now I see there's no `prepare_merykitty*.sh`.

src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java
0x3C50: lambda bootstrapping takes a short while once, but since invokedynamic is involved, each subsequent lambda "creation" takes the short path. anonymous inner classes are probably slightly faster, but it won't really matter here since that's just a few nanos at the start of the program. if you're doing this kind of optimization, you're using the wrong language

franz1981: I don't agree on the language choice. Read the assembly, and if there is a way to achieve it, just do it. Languages are just tools and the same for compilers; it's different in case the effort to achieve the effect is too much, but in this case, is a one liner.

PR #339, opened 2024-01-11
Use bash in prepare scripts

AlexanderYastrebov:
Prepare scripts that use `#!/bin/sh` and
`source "$HOME/.sdkman/bin/sdkman-init.sh"`
fail on systems where `sh` is not `bash`,
e.g. on Ubuntu it is `dash` which has not `source`.


gunnarmorling:
Thx!

PR #340, opened 2024-01-12
Performance improvements for CalculateAverage_ricardopieper

ricardopieper:
#### Check List:
* Execution time: 9s
* Execution time of previous implementation: 11.2s

I determined that Linux is a bit faster with mmap reads than my mabook. On my Windows PC it's significantly faster. I still want it to run fast on my mac, so I developed a mechanism to switch file readers based on the OS. I don't know if my macbook has something misconfigured or not because this is not universal. In fact Roy Van Rijn's macbook does extremely fast mmap reads.

Also, as I suspected the final merge was taking a whole second too much. More like 2 seconds...

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,7 +11,7 @@ Thanks for your submission. Please go through the checklist above before submitt
 Use [x] to mark that the item has been completed.
 
 Due to the large number of entries created so far,
-please submit only entries that are you are expecting to run in 30 seconds or less.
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hum, this is significantly slower than what's on main right now. Is actually faster on your machine? Also, can you please rebase this and squash it into a single commit off of latest main? Thx.

ricardopieper:
I will reset my branch and open a new one. Afraid of screwing things up.

PR #342, opened 2024-01-12
Add Windows support

kant2002:
this is parts of the scripts required to run validation on Windows. I can work on other scripts if there is interest in that.

### DIFF
--- a/calculate_average_baseline.cmd
+++ b/calculate_average_baseline.cmd
@@ -1,0 +1,46 @@
+@echo off
+@setlocal
+REM !/bin/sh
+REM 
+REM   Copyright 2023 The original authors
+REM 
+REM   Licensed under the Apache License, Version 2.0 (the "License");
+REM   you may not use this file except in compliance with the License.
+REM   You may obtain a copy of the License at
+REM 
+REM       http://www.apache.org/licenses/LICENSE-2.0
+REM 
+REM   Unless required by applicable law or agreed to in writing, software
+REM   distributed under the License is distributed on an "AS IS" BASIS,
+REM   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+REM   See the License for the specific language governing permissions and
+REM   limitations under the License.
+REM 
+
+REM  Uncomment below to use sdk
+REM  source "$HOME/.sdkman/bin/sdkman-init.sh"
+REM  sdk use java 21.0.1-graal 1>&2
+
+SET JAVA_OPTS=
+set start=%time%
+
+java %JAVA_OPTS% -classpath target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage
+
+set end=%time%
+set options="tokens=1-4 delims=:.,"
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Hey, thx a lot for creating this. I have to admit, I'm a bit reluctant here, we don't run evaluation on Windows, and I don't even have access to such a machine. Can't Windows run Shell scripts too these days via WSL? Or people could use cygwin, if that's still a thing. I'd prefer we'd not duplicate this infrastructure and use the same scripts everywhere.

gunnarmorling:
Hey @ddimtirov, IIRC, you are on Windows too. Do you think adding this would be beneficial? I'd prefer if we wouldn't have to maintain two sets of these scripts. Has been many years since I've been working with a Windows machine, but I think there are means of running shell scripts reasonably well=

ddimtirov:
I switched to WSL, as I'd like to be able to run the other submissions, and some of them fail on Windows (line ending issues, and possibly others). I would recommend others do the same - it is not difficult, especially if you are running IDEA Ultimate.

The scripts above are OK, though I'd normally use Powershell, and [`Measure-Command`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/measure-command?view=powershell-7.4) instead of the sophisticated time calculations. 

Anyone using Git for Windows has got a functioning `bash` shell, that can run the `calculate_...` scripts and `create_measurements`, though I don't think they would br ablr to `test_...` unless we replace the `ln -s` with copying (the files are small enough, and the MinGW's `ln` does not create real NTFS symlinks).

gunnarmorling:
Thx for weighing in. In that light, I am going to close this one without merging. It seems there are alternatives for folks running on Windows, and as I don't have that environment, it's too cumbersome to maintain those extra scripts. Thx a lot for looking into this nevertheless!

ddimtirov:
I am going to address two of the arguments above:

> But I honestly do not buy that installing WSL or Cygwin is solution. 

These day and age, I also would stay away from Cygwin, but Mingw and WSL have their place. 

> now I should have two similar toolchains on my computer which is not very pleasant. 

You can install Git, Zip and Unzip using your distro packages, and then SDKMan in total 15 minutes.

That's all you need - with SDKMan it is trivial to experiment with different JDKs, the project uses Maven Wrapper, so you don't need to worry about that.

Downsides are that you may need to do some learning - how to use WSL, how to use your IDE with WSL backend (I'd avoid installing IDE inside WSL, but the newer versions support Wayland)

Another downside is that it takes disk space - it is cheap, but not free. This is the reason I spun up a new "distribution" for this project that I'm going to delete after I'm done. Because of this I was also able to use the exact same Fedora version that Gunnar uses in his setup.

> WSL is virtualized environment for running Linux which is even farther for running even something approximately the same.

In fact, WSL being virtualized environment brings it closer to the testing setup - a VM on the Hetzner cloud.

gunnarmorling:
> In fact, WSL being virtualized environment brings it closer to the testing setup - a VM on the Hetzner cloud.

That's actually not the case any more. We've moved to a dedicated machine (Hetzner AX161), as the original VM got relocated to a different host and thus numbers weren't comparable any longer. The leaderboard has been updated to reflect that change.

PR #349, opened 2024-01-12
Performance improvements for ricardopieper

ricardopieper:
#### Check List:
* Execution time: 9s
* Execution time of previous implementation: 11.2s

I determined that Linux is a bit faster with mmap reads than my mabook. On my Windows PC it's significantly faster. I still want it to run fast on my mac, so I developed a mechanism to switch file readers based on the OS. I don't know if my macbook has something misconfigured or not because this is not universal. In fact Roy Van Rijn's macbook does extremely fast mmap reads.

Also, as I suspected the final merge was taking a whole second too much. More like 2 seconds...

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_ricardopieper.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_ricardopieper.java
@@ -23,6 +23,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
+
+import java.time.Duration;
+import java.time.Instant;
 import java.util.*;
 
 /*
@@ -36,6 +39,8 @@ import java.util.*;
   - @royvanrijn again for the parse int only and /10 in the very end
   - @flippingbits for the idea of not calling getLong on mmaped file directly, get a big chunk into memory instead
         - Maybe on linux it's actually faster?
+        - Confirmed: mmap is faster at least on Windows.
+
   - OpenAI / ChatGPT for the actual bit twiddling idea for finding a character, it couldn't explain it so I did my best
     (Turns out @royvanrijn had the same idea)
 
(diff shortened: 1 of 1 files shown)


ricardopieper:
@gunnarmorling in the previous commit you mentioned the performance of this new version is much lower. Could it be due to disk cache?

I am running this on an amazon machine c6a.4xlarge with io2 60GB ssd (provisioned iops) and as you can see the first execution was slower, but then it became ~9 seconds in the following ones.

Screenshot 2024-01-12 at 11 41 40 AM

gunnarmorling:
I run everything from a RAM disk, so definitely not disk caching issue. Hard to tell what's going without digging into the details. This is what I get as of this PR:

```
Benchmark 1: timeout -v 300 ./calculate_average_ricardopieper.sh 2>&1
  Time (mean ± σ):      9.754 s ±  0.086 s    [User: 74.182 s, System: 0.760 s]
  Range (min … max):    9.625 s …  9.835 s    5 runs

Summary
  ricardopieper: trimmed mean 9.770500059566666, raw times 9.8143185969,9.7858196979,9.6247024009,9.7113618839,9.8354425609

Leaderboard

|   | 00:09.770 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_ricardopieper.java)| 21.0.1-graal | [Ricardo Pieper](https://github.com/ricardopieper) |  |
```

I suppose something related to the specific ISA and its extensions (Zen2), insights e.g. from Mac don't transfer at all.

ricardopieper:
Maybe it's due to a new allocation in the main loop... but that's pretty odd. I will try making a ramdisk and see what i can do.

gunnarmorling:
Hey @ricardopieper, I'm gonna close this one for now as I haven't heard back for a while and I need to clean the slate a bit. Please feel free to reopen once you feel there's an update which improves perf here. Thanks a lot!

PR #350, opened 2024-01-12
Bug: Update class name in calculate_average_baseline.sh

gnabyl:
The baseline implementation has been renamed to `CalculateAverage_baseline`


gunnarmorling:
Ah yes, good catch. Thx!

PR #352, opened 2024-01-12
Locally another ~3% faster, much faster for larger set, made more general

royvanrijn:
#### Check List:
* Execution time: `1.305s` locally, before was `1.340s`
* Execution time of reference implementation: -

Added a lot of comments for when people want to read and learn more.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
@@ -23,7 +23,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -53,8 +52,13 @@ import sun.misc.Unsafe;
  * Various tweaks for Linux/cache    1550 ms (should/could make a difference on target machine)
  * Improved layout/predictability:   1400 ms
  * Delayed String creation again:    1350 ms
+ * Remove writing to buffer:         1335 ms
+ * Optimized collecting at the end:  1310 ms
+ * Adding a lot of comments:         priceless
  *
  * Big thanks to Francesco Nigro, Thomas Wuerthinger, Quan Anh Mai for ideas.
+ *
+ * Follow me at: @royvanrijn
  */
 public class CalculateAverage_royvanrijn {
 
@@ -74,29 +78,24 @@ public class CalculateAverage_royvanrijn {
     }
 
     public static void main(String[] args) throws Exception {
-
         // Calculate input segments.
         final int numberOfChunks = Runtime.getRuntime().availableProcessors();
         final long[] chunks = getSegments(numberOfChunks);
 
-        final List<Entry[]> repositories = IntStream.range(0, chunks.length - 1)
+        final Map<String, Entry> measurements = HashMap.newHashMap(1 << 10);
+        IntStream.range(0, chunks.length - 1)
                 .mapToObj(chunkIndex -> processMemoryArea(chunks[chunkIndex], chunks[chunkIndex + 1]))
                 .parallel()
-                .toList();
-
-        // Sometimes simple is better:
-        final HashMap<String, Entry> measurements = HashMap.newHashMap(1 << 10);
-        for (Entry[] entries : repositories) {
-            for (Entry entry : entries) {
-                if (entry != null)
-                    measurements.merge(extractedCityFromLongArray(entry.data, entry.length), entry, Entry::mergeWith);
-            }
-        }
+                .forEachOrdered(repo -> { // make sure it's ordered, no concurrent map
+                    for (Entry entry : repo) {
+                        if (entry != null)
+                            measurements.merge(turnLongArrayIntoString(entry.data, entry.length), entry, Entry::mergeWith);
+                    }
+                });
 
         System.out.print("{" +
                 measurements.entrySet().stream().sorted(Map.Entry.comparingByKey()).map(Object::toString).collect(Collectors.joining(", ")));
         System.out.println("}");
-
     }
 
     /**
@@ -123,15 +122,20 @@ public class CalculateAverage_royvanrijn {
         }
     }
 
-    private static final int TABLE_SIZE = 1 << 19; // large enough for the contest.
-    private static final int TABLE_MASK = (TABLE_SIZE - 1);
-
+    // This is where I store the hashtable entry data in the "hot loop"
+    // The long[] contains the name in bytes (yeah, confusing)
+    // I've tried flyweight-ing, carrying all the data in a single byte[],
+    // where you offset type-indices: min:int,max:int,count:int,etc.
+    //
+    // The performance was just a little worse than this simple class.
     static final class Entry {
-        private final long[] data;
-        private int min, max, count, length;
-        private long sum;
 
-        Entry(final long[] data, int length, int temp) {
+        private int min, max, count;
+        private byte length;
+        private long sum;
+        private final long[] data;
+
+        Entry(final long[] data, byte length, int temp) {
             this.data = data;
             this.length = length;
             this.min = temp;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -164,127 +168,161 @@ private static double round(double value) {
...
 
+    private static final long SEPARATOR_PATTERN = 0x3B3B3B3B3B3B3B3BL;
+
+    // Takes a long and finds the bytes where this exact pattern is present.
+    // Cool bit manipulation technique: SWAR (SIMD as a Register).
     private static long getDelimiterMask(final long word) {
-        long match = word ^ SEPARATOR_PATTERN;
-        return (match - 0x0101010101010101L) & ~match & 0x8080808080808080L;
+        final long match = word ^ SEPARATOR_PATTERN;
+        return (match - 0x0101010101010101L) & (~match & 0x8080808080808080L);
```
royvanrijn, line 309: This was so obvious, but had not yet crossed my mind. I thought the compiler would check that.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -74,29 +78,24 @@ private static Unsafe initUnsafe() {
...
-
         // Calculate input segments.
         final int numberOfChunks = Runtime.getRuntime().availableProcessors();
         final long[] chunks = getSegments(numberOfChunks);
 
-        final List<Entry[]> repositories = IntStream.range(0, chunks.length - 1)
+        final Map<String, Entry> measurements = HashMap.newHashMap(1 << 10);
+        IntStream.range(0, chunks.length - 1)
                 .mapToObj(chunkIndex -> processMemoryArea(chunks[chunkIndex], chunks[chunkIndex + 1]))
                 .parallel()
```
pivovarit, line 88: Have you tried manual parallelization with an array of `Thread`s?

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -164,127 +168,161 @@ private static double round(double value) {
...
     /**
      * For case multiple hashes are equal (however unlikely) check the actual key (using longs)
      */
-    static boolean arrayEquals(final long[] a, final long[] b, final int length) {
-        for (int i = 0; i < length; i++) {
-            if (a[i] != b[i])
+    private static boolean memoryEqualsEntry(final long startAddress, final long[] entry, final long finalBytes, final int amountLong) {
+        for (int i = 0; i < (amountLong - 1); i++) {
+            int step = i << 3; // step by 8 bytes
+            if (UNSAFE.getLong(startAddress + step) != entry[i])
```
thomaswue, line 320: I think the compiler might not be able to proof that (i < entry.length) here at all times and so maybe also unsafe is a minor improvement for the array access here. Once you go unsafe, you have to be all in ;-).

I kind of liked that buffer array before btw, and was surprised by how fast this worked.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
SamuelYvon: I'm not familiar with all the details of @royvanrijn's implementation, but I imagine that if you have more chunks than (system) threads available (mine implementation may, if the chunks are too large to all fit within a `mmap`) you might have parked threads. I _assume_ `parallel` uses work-stealing when a thread is done. I'm not sure how java manages the virtual threads, so this entire comment may be irrelevant

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -53,8 +52,13 @@
  * Various tweaks for Linux/cache    1550 ms (should/could make a difference on target machine)
  * Improved layout/predictability:   1400 ms
  * Delayed String creation again:    1350 ms
+ * Remove writing to buffer:         1335 ms
+ * Optimized collecting at the end:  1310 ms
+ * Adding a lot of comments:         priceless
```
SamuelYvon, line 57: ❤️

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
```diff
@@ -164,127 +168,161 @@ private static double round(double value) {
...
+
+    // Takes a long and finds the bytes where this exact pattern is present.
+    // Cool bit manipulation technique: SWAR (SIMD as a Register).
     private static long getDelimiterMask(final long word) {
-        long match = word ^ SEPARATOR_PATTERN;
-        return (match - 0x0101010101010101L) & ~match & 0x8080808080808080L;
+        final long match = word ^ SEPARATOR_PATTERN;
+        return (match - 0x0101010101010101L) & (~match & 0x8080808080808080L);
+        // I've put some brackets separating the first and second part, this is faster.
+        // Now they run simultaneous after 'match' is altered, instead of waiting on each other.
```
gunnarmorling, line 311: Interesting. Out of curiosity, how do you know that?

gunnarmorling:
00:02.855 now, i.e. not a huge difference.

```
  Time (mean ± σ):      2.856 s ±  0.012 s    [User: 19.831 s, System: 0.701 s]
  Range (min … max):    2.836 s …  2.875 s    10 runs

Summary
  royvanrijn: trimmed mean 2.85591370276, raw times 2.86659514426,2.84911410126,2.83607515926,2.83736008126,2.85235688026,2.8585244732599997,2.85607449126,2.86436423126,2.86292021926,2.8747086562599997

Leaderboard

|   | 00:02.855 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java)| 21.0.1-graal | [Roy van Rijn](https://github.com/royvanrijn) | GraalVM native binary |
```

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
royvanrijn: Running it with and without the brackets, at this point I'm just changing things and hoping for the best.

It does make sense though, first calculate (match - something) and (~match & something) without a dependency, than AND together.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
royvanrijn: It doesn't, I've tried submitting parts and starting to process once the CompletableFuture is done; but all of that adds overhead more than it solves a problem.

The parallel() can't really work-steal, the ranges are set by the chunk array.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
royvanrijn: I've already done just that; I've been keeping two branches locally with much more Unsafe; and until just now they were slower on my machine. However, with the latest changes, it is now actually faster, so I'm switching over to more and more Unsafe method calls and doing only memory.

I do still copy the bytes into the table, I think this will improve speed for the larger dataset specifically.

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
balayanv: what if you use counted completers, your main thread will not split to N segments but start walking trough the data to identify separator, each time it identifies other workers can steal and process the entry... once main is done identifying all of the entries it joins and consumes whatever is left.

Basically splitting to N chunks (which are not really the same size +- 2 entires) means your slowest worker is the number you get. Since in theory you are doing the walk anyway and identifying the separators you are not doing any extra work but using true work steal... 

You can probably improve the speed of seeking the separators if yu know how many bytes you can confidently skip after last separator... the name and whatever boilerplate is in the file will give you some offset that you can skip reliably. I'm not sure if names have minimum length etc..

balayanv: though i will assume the volatile write that happens to get things synchronised will cost you more than you benefit from perfectly distributing work...

src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
balayanv: just checking if i understand this correctly... there is no parallelisation here right? the reason (1 - 2) & (~3 &4) is faster is because things end up in correct registers so the last & is faster?

mtopolnik: Roy can correct me, but I think he means that It's parallel within the CPU thanks to its superscalar hardware. Even a single core has multiple ALUs.

balayanv: thanks, in that case its indeed surprising that this is not done on the compiler level, pretty easy to detect...

PR #353, opened 2024-01-12
Update create_fork.sh

ivanklaric:
Not creating a submission yet, I just found that create_fork .sh doesn't work since you renamed CalculateAverage to CalculateAverage_baseline, so I thought I'd fix that first.


gunnarmorling:
Good catch. Updated that earlier today and seems I've missed to adjust it in a couple places. Thx!

PR #356, opened 2024-01-12
[Attempt #2] String overflow hash + data/mem optimization

gnabyl:
#### Check List:
* Execution time: 0m6.434s
* Execution time of reference implementation: > 4m

`evaluate2.sh` gives close runtime compare to submissions that run under 10 seconds. So I would love to give it a try, since it's quite hard to replicate the same evaluation env.

@gunnarmorling could you run this without merging?

### DIFF
--- a/calculate_average_gnabyl.sh
+++ b/calculate_average_gnabyl.sh
@@ -19,5 +19,5 @@
 # source "$HOME/.sdkman/bin/sdkman-init.sh"
 # sdk use java 21.0.1-graal 1>&2
 
-JAVA_OPTS=""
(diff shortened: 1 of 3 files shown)


gunnarmorling:
~8 sec now. Also remove the `time` call from the launch script, it will safe a few ms.

gnabyl:
@gunnarmorling 
Thanks for the suggestion, I think this PR is ready to be merged. I wanted to try something but it didn't work out. I'll have to sleep on it a bit later.

gunnarmorling:
00:08.489 now. Thx for participating!

PR #358, opened 2024-01-12
Improve scheduling for thomaswue

thomaswue:
#### Check List:
* Execution time: 0.62

Already filing enhancement requests for the GraalVM compiler scheduler. Apparently this innocent looking change gains 6%+ ;-).


gunnarmorling:
It doesn't seem to move the needle on the eval machine:

```
Benchmark 1: timeout -v 300 ./calculate_average_thomaswue.sh 2>&1
  Time (mean ± σ):      2.722 s ±  0.008 s    [User: 18.719 s, System: 0.701 s]
  Range (min … max):    2.716 s …  2.741 s    10 runs

Summary
  thomaswue: trimmed mean 2.72014917887, raw times 2.71617197462,2.7215708666199996,2.71872755162,2.7185243566199997,2.7305528786199997,2.7414847446199997,2.72071959562,2.71618575162,2.7173498986199998,2.7175625316199996

Leaderboard

|   | 00:02.720 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java)| 21.0.1-graal | [Thomas Wuerthinger](https://github.com/thomaswue) | GraalVM native binary |
```

gunnarmorling:
> Already filing enhancement requests for the GraalVM compiler scheduler.

This is great! Also someone mentioning JDK tickets somewhere. So great to see these improvements as a fallout of this!

gunnarmorling:
Yepp, absolutely. It might be as far as we get for now, which is waaay beyond whatever I had been expecting before. Let's see whether someone manages to pull out a surprise bunny out of their hat between now and Jan 31 ;)

thomaswue:
@gunnarmorling No stress evaluating this, but now I fixed a problem in the hash calculation (which even led to a collision for the test data set) and improved the performance for the new 10k cities test substantially by being a bit better in case of actual collisions. The performance gain on my machine was another 3% over the previous version for the original test data and I have now runs slightly below 0.60s there.

gunnarmorling:
00:02.552 now on the official data set. I.e. a teenie-tiny bit ahead @merykitty's recorded unsafe entry. So close in fact that I can't identify a clear ordering with the current way of measuring. I'm therefore putting both with #1 into the leaderboard for now. I don't think the order matters an awful lot, as the learning experience is the most important outcome here.

```
| 1 | 00:02.552 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java)| 21.0.1-graal | [Thomas Wuerthinger](https://github.com/thomaswue), [Quan Anh Mai](https://github.com/merykitty), [Alfonso² Peterssen](https://github.com/mukel) | GraalVM native binary |
```

thomaswue:
Yes, agreed, thank you!

PR #363, opened 2024-01-13
Rename files to match GitHub username

as-com:
Seems the script that updates the leaderboard linked to the wrong asun. This should fix it.


gunnarmorling:
Yeah, names are resolved based on the launcher script name. Thx for the update!

PR #366, opened 2024-01-13
added python script to build test data

Eadrom:
Hello,

The challenge sounded like it would be fun. I don't know java or have a development environment setup, but I thought it would be fun to try it out in python and go. I'd need test data to get started but the code to generate that is java. I figure if anyone else is wanting to try the challenge, but not in java, they probably have python available, especially on a *nix box. So I thought I would offer this in case others looking to do the challenge, but not in java, might find it useful. 

It only uses builtin modules, no other libraries need to be installed. It should dump the test data into the same directory as the java counterpart. I did my best to make sure that it generates test data in the same format as the java CreateMeasurements with the same limitation on a max of 10,000 unique station names. 

If you rather not include it in a java-centric challenge's repo, I completely understand. I just thought I would offer. 

Thanks for the fun challenge idea!

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ Execute the following steps to run the challenge:
     This will take a few minutes.
     **Attention:** the generated file has a size of approx. **12 GB**, so make sure to have enough diskspace.
 
+    If you're running the challenge with a non-Java language, there's a non-authoritative Python script to generate the measurements file at `src/main/python/create_measurements.py`. The authoritative method for generating the measurements is the Java program `dev.morling.onebrc.CreateMeasurements`.
+
 3. Calculate the average measurement values:
 
     ```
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Nice, we can add it, seems easy enough. Can you add a note to the relevant part of the README (clarifying that the Java-based generator is the authoritative one) and move the file to src/main/python? Thx!

Eadrom:
@gunnarmorling I believe I've sorted out what you asked for. Please let me know if there's anything else to change or update. 

Thank you!

gunnarmorling:
Nice one, thx!

harenbrs:
I'd like to note that this Python script generates data with a very different distribution (uniform from -99.9 to 99.9 for every station) from the original Java version (normally distributed around real mean temperature values for each station with standard deviation of 10).

Edit: Forgot to mention, it also uses way more station names (choosing from ~10k samples from `weather_stations.csv`) than the original Java version (412 hardcoded). Arguably this difference is more algorithmically significant.

PR #367, opened 2024-01-13
CalculateAverage_ivanklaric

ivanklaric:
#### Check List:
- [x] Your launch script is named `calculate_average_ivanklaric.sh` (make sure to match casing of your GH user name) and is executable
* Execution time:  01:11
* Execution time of reference implementation: 02:45

I haven't coded in Java since 1.6 days, so I learned a ton about Stream API by shamelessly being inspired by CalculateAverage_abfmblr.java. Tweaked it a bit to get minor (arguably within measurement error) improvements by not using String.split(), Might go one step further to see if parseDouble() can be replaced with something faster, but I think this is a decent first attempt :)

### DIFF
--- a/calculate_average_ivanklaric.sh
+++ b/calculate_average_ivanklaric.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Note that as per the README and the PR template, new entries at this point should run in 10 sec or less, as it's just too much to handle otherwise with the large number of submissions so far. This one comes in at around 01:25.801, i.e. way longer than that. I'm still gonna merge it, as you've already put in the effort. But should you plan to keep working on it, it would be good to get into the required range. Happy to hear that this was interesting to you, thanks for participating!

PR #373, opened 2024-01-13
Leaderboard: Update my time & position after #356

gnabyl:
Update my time & position after #356. Just to help out a bit since it's a lot of manual work for you


gunnarmorling:
Thx. I apply them in batches, that's why it wasn't reflected here yet.

Issue #49, opened 2024-01-03
Clarify rounding semantics


gunnarmorling:
See https://github.com/gunnarmorling/1brc/pull/5#discussion_r1440578784; CC @AlexanderYastrebov

ddimtirov:
I find that `Math.round((double) x / y) / 10.0` produces the same result as the reference - the important part is to do the division in `double` - addition is guaranteed to be loseless.

Also, I have redone the reference impl as BigDecimal with rounding half-even, and the results don't change (for my file).

[CalculateAverage_BigDecimal.java.txt](https://github.com/gunnarmorling/1brc/files/13828834/CalculateAverage_BigDecimal.java.txt)

gunnarmorling:
Done via #392.

AlexanderYastrebov:
> Done via https://github.com/gunnarmorling/1brc/pull/392.

Unfortunately no, see https://github.com/gunnarmorling/1brc/pull/392/files#r1451734941

PR #375, opened 2024-01-13
Compare outputs with tolerance

mtopolnik:
This adds a Java program that compares two output files with tolerance for accumulated `double` errors.

It also produces nice diagnostic output, indicating the whole entry where there's a problem.


src/main/java/dev/morling/onebrc/CompareWithTolerance.java
```diff
@@ -0,0 +1,105 @@
...
+package dev.morling.onebrc;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+import java.util.stream.IntStream;
+
+public class CompareWithTolerance {
```
gunnarmorling, line 25: Can we make this an implicit class in the root dir, with top-level main method, like `process_output.java` (which itself btw. is obsolete by now, I'm gonna remove it). That way, it doesn't have to be re-packaged after changes, just by running it with `java lenient_compare.java`.

src/main/java/dev/morling/onebrc/CompareWithTolerance.java
```diff
@@ -0,0 +1,105 @@
...
+    private static final Pattern numRegex = Pattern.compile("-?\\d\\d?\\.\\d");
+
+    record NameAndStats(String name, int[] stats) {
+    }
+
+    public static void main(String[] args) throws IOException {
+        var fname1 = args[0];
+        var fname2 = args[1];
```
gunnarmorling, line 33: Rather 1 and 2, let's use "expected" and "actual", to clarify which is which.

src/main/java/dev/morling/onebrc/CompareWithTolerance.java
```diff
@@ -0,0 +1,105 @@
...
+        var entries1 = entries(fname1);
+        var entries2 = entries(fname2);
+        if (entries1.length != entries2.length) {
+            System.out.printf("Files don't match in the number of entries: %,d vs. %,d%n", entries1.length, entries2.length);
+            System.exit(1);
+        }
+        for (int i = 0; i < entries1.length; i++) {
+            var nameAndStats1 = parseEntry(entries1[i], i, fname1);
+            var nameAndStats2 = parseEntry(entries2[i], i, fname2);
+            if (IntStream.range(0, 3)
```
gunnarmorling, line 43: Leniency should only be applied to the mean value, I don't think the current baseline would get min or max wrong?

src/main/java/dev/morling/onebrc/CompareWithTolerance.java
```diff
@@ -0,0 +1,105 @@
...
+                    .mapToObj(n -> new double[]{ nameAndStats1.stats[n], nameAndStats2.stats[n] })
+                    .anyMatch(pair -> Math.abs((pair[0] - pair[1])) > 1)) {
```
gunnarmorling, line 45: I think the baseline would only ever get the value too high by 0.1 (or 1 as `int` here), i.e. no need for `abs()`?

src/main/java/dev/morling/onebrc/CompareWithTolerance.java
```diff
@@ -0,0 +1,105 @@
...
+                System.out.printf("Values don't match in entry #%,d: \"%s\" vs. \"%s\"%n", i, entries1[i], entries2[i]);
+                System.exit(1);
+            }
+        }
+    }
+
+    private static String[] entries(String fname) throws IOException {
+        var content = new BufferedReader(new FileReader(fname)).readLine();
```
gunnarmorling, line 53: Should be in try-with-resource for good measure.

test.sh
```diff
@@ -45,7 +45,7 @@ for sample in $(ls $INPUT); do
   rm -f measurements.txt
   ln -s $sample measurements.txt
 
-  diff <("./calculate_average_$FORK.sh" | ./tocsv.sh) <(./tocsv.sh < ${sample%.txt}.out)
+./compare_output.sh <("./calculate_average_$FORK.sh") ${sample%.txt}.out
```
gunnarmorling, line 48: Can we make this parameterized instead, i.e. add a `--lenient` option to the script which controls which mode to use?

gunnarmorling:
Dang, all this is a bit of a mess. Turns out, I even had the correct behavior initially (see commented code in baseline), but then I got "smart" and refactored it into that `round()` method. Someone brought up the concern about missing creating false positives with this approach (i.e. masking actual errors). Seems kinda impossible to not make it suck in one way or another. Maybe, the least bad way could be this:

* Fix the baseline
* Allow tests to be run with two modes: "strict" (i.e. current diff-based impl, contenders must exactly match the result of the fixed baseline) and "lenient" (i.e. what's proposed here, contenders can under-report mean by 0.1 compared to the fixed baseline)
* Leave existing entries as is (i.e. implicitly using "lenient" validation mode)
* Use "strict" mode for new or updated entries going forward
* When publishing the 10K key set test results for the 10 or so top entries (i.e. similar to your earlier experiment), add a column which says whether they pass in "lenient" or "strict" mode, encouraging entries which only pass in "lenient" mode to adjust their implementation

I feel that may be the best balance to ensure correctness while not moving goal posts too much at this point. In particular, it would not mess up any existing results so far, which feels important to me.

@mtopolnik, @royvanrijn, @hundredwatt, @thomaswue, @merykitty, any thoughts?

gunnarmorling (commented):
Thanks! A few comments inline.

mtopolnik:
> @mtopolnik, @royvanrijn, @hundredwatt, @thomaswue, @merykitty, any thoughts?

Maybe we can first assess the damage? How many solutions are failing? For example, with my dataset, I found only a single solution to have this off-by-one-decimal problem.

royvanrijn:
Sounds like a good plan. If there are entries failing, let’s mark that separately on the leaderboard and they’ll hopefully be fixed.

mtopolnik:
I can add to the script that it records such minor mistakes, but continues testing for larger ones. Then we know for sure the solution is correct module off-by-decimal.

tivrfoa:
Maybe an easier approach to not have to change testing too much:

- Copy `CalculateAverage_baseline` to `CalculateAverage_baseline_lenient`
- Copy `CalculateAverage_baseline` to `CalculateAverage_baseline_strict`
- Fix `strict` version
- Use `diff` first on strict output, then on lenient output if needed.

gunnarmorling:
> Maybe we can first assess the damage? How many solutions are failing? For example, with my dataset, among the first 20 I found only a single solution to have this off-by-one-decimal problem.

For the test case brought up by @tivrfoa [here](https://twitter.com/tivrfoa/status/1746134085959819540), about 1/3 of _all_ implementations fail when comparing to the output of the fixed baseline. But out of the current Top 20, i.e. the set I'd run for the 10K keyset potentially, only 2 fail. So not too bad indeed.

gunnarmorling:
> Maybe an easier approach to not have to change testing too much...

Yeah, coming to the same conclusion actually. So I think just adding the new test case should be all we need here, and I'll follow up with updating the baseline implementation, keeping the old one around two for cases where the old output needs to be reproduced.

AlexanderYastrebov:
This seems to address #49

validate_output.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private static String[] entries(String fname) throws IOException {
+        var content = new BufferedReader(new FileReader(fname)).readLine();
+        if (!content.startsWith("{") || !content.endsWith("}")) {
+            System.out.printf("%s isn't in the format \"{entries}\"%n", fname);
+            System.exit(1);
+        }
+        content = content.substring(1, content.length() - 1);
+        return content.split(",");
```
AlexanderYastrebov, line 78: The whole thing kind of shows the importance of using a proper format https://github.com/gunnarmorling/1brc/issues/14

See also https://github.com/gunnarmorling/1brc/blob/a0f826c2e4441c93b4c6cf54df77793c03f2cc67/tocsv.sh#L29-L31

gunnarmorling:
Ok, so as discussed I'm moving forward with this by only adding the new test case and adjusting the baseline implementation accordingly. I'm going to close this PR in favour of https://github.com/gunnarmorling/1brc/pull/392, which contains this work.

New or updated entries need to be compliant with this fixed behavior, i.e. pass the new additional test, whereas existing entries remain as-is, i.e. compliance at the point of original evaluation is relevant. Thanks a lot @mtopolnik for pushing on this, and to all for providing feedback.

mtopolnik:
Here's the summary of what I understood. Many solutions do this: accumulate an integer sum that is 10x the actual value, and then, to get the average:

1. Convert both sum and count to double;
2. Divide sum by count;
3. Round the result;
4. Divide by ten.

The baseline accumulates the sum as a double, and then:

1. Divide sum by count;
2. Multiply by 10.0;
3. Round;
4. Divide by ten.

The proposed fix is to instead:

1. Multiply the sum by 10.0;
2. Round;
3. Divide by 10.0;
4. Divide by count;
5. Round again.

This should align the result with the first approach, which accumulates the integers.

gunnarmorling:
Yes, pretty much that. See the linked new PR for how it looks like from a "specification" point of view.

PR #376, opened 2024-01-13
Simultaneous reading and calculation, reduction of memory consumption…

jgrateron:
simultaneous reading and calculation, reduction of memory consumption, hashing for the stations
Test pass
formating
add coments spanish
Execution time: 00:17 sec before 00:28 sec


gunnarmorling:
00:18.313 now. Thx for participating!

PR #388, opened 2024-01-14
A SAFE and readable version

anitasv:
#### Check List:
* Execution time: 3.741 s
* Execution time of reference implementation:  3 minutes, 18 seconds

From the same team (@vaidhy  and @anitasv ) that submitted a faster one (2.4 seconds on our machine). This a full rewrite, without any UNSAFE.* functions or complex bit manipulations, getLong() for speed ups etc. Idea being something someone might naturally write and maintain, but luckily just happened to be really fast :).

Actually removing the custom HashMap with java builtin HashMap increases only by 1 second. If I can find a way to avoid it, I will submit another correction.

### DIFF
--- a/calculate_average_anitasv.sh
+++ b/calculate_average_anitasv.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


anitasv:
Removed custom hash map also, now it is 200 milliseconds slower. 3.9 seconds.

gunnarmorling:
00:08.752, nice!

PR #390, opened 2024-01-14
Compare results with tolerance

AlexanderYastrebov:
As an exercise I've re-implemented #375 idea to compare result numbers with a tolerance:
```
$ ./test.sh baseline 0 src/test/resources/samples/measurements-rounding-precise.txt
Validating calculate_average_baseline.sh -- src/test/resources/samples/measurements-rounding-precise.txt
Rounding=14.6/25.5/33.6 != Rounding=14.6/25.4/33.6 (avg)
```
The hassle of parsing highlights the importance of using machine-readable format as suggested in #14

**But** I think this approach is wrong, the rules should define rounding mode and baseline should be fixed to produce the correct result.

Submissions should be fixed accordingly or qualified as non-passing.

Could have been done earlier but better late than never.

Updates #49


gunnarmorling:
Hey @AlexanderYastrebov, yeah, agreed that the "compare with tolerance" approach isn't the best one. I have opened https://github.com/gunnarmorling/1brc/pull/392 which fixes the baseliine and adds a test case to ensure the correct behavior. Note I am not going to re-evaluate existing entries, compliance at the time of evaluation is what decides (it's a common strategy, for instance also used in JCP when certifying spec implementations). New or updated entries must be compliant with the fixed behavior, by passing the additional test.

gunnarmorling:
So do we still need this in the light of the above, @AlexanderYastrebov?

AlexanderYastrebov:
> So do we still need this in the light of the above

No, this was just an exercise.

PR #106, opened 2024-01-04
#104 Run tests for the user creating the PR

filiphr:
Adding a step in the GitHub action to test the submission of the user submitting the PR

Fixes #104

@gunnarmorling please wait for merging, I want to also push some failures to see how it looks like.

cc @AlexanderYastrebov

Issue #104, opened 2024-01-04
Run test suite as part of PR build


gunnarmorling:
Provided folks stick to the requested convention and name their launch script after their user name, we should be able to run the _tests.sh_ script as part of PR builds and fail them accordingly.

CC @AlexanderYastrebov, @filiphr

PR #393, opened 2024-01-14
#104 Run tests for the user creating the PR and validate that the calculate file exists for the user

filiphr:
Also install all available SDks before running the tests

@gunnarmorling this is a continuation of PR #106 where I am also adding the caching and installation of all available SDKs. If a new SDK is needed it would need to be added to the `available_sdks.txt` file.

Let me know what you think


filiphr:
Seems like the sdkman install fails due to the interactive mode. I'll investigate a bit more and see if I can somehow disable it.

PR #398, opened 2024-01-14
karthikeyan97 implementation

karthikeyan97:
#### Check List:
* Execution time:4.5
* Execution time of reference implementation:208

### DIFF
--- a/calculate_average_karthikeyan97.sh
+++ b/calculate_average_karthikeyan97.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


prepare_karthikeyan97.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
gunnarmorling, line 1: That file is not needed when empty.

gunnarmorling:
Hey, thx for particpating! As per the README and the PR template, new entries should run under 10 sec at this point, as it's just getting to hard to operate otherwise. It's not a hard and fast rule, but this one comes in at 39.995 s, i.e. way beyond the threshold. So maybe move it to draft state for now and iterate a bit on your end until it's there? Thx!

gunnarmorling (changes requested):
See above.

karthikeyan97:
Hi made some changes and think this will do the magic.Can u please run this once again

PR #413, opened 2024-01-15
CalculateAverage_gonix initial attempt

gonix:
#### Check List:
* Execution time: ~2.5s
* Execution time of reference implementation: ~145s
* Machine: Intel i7-11800H @ 2.30GHz (max 4.6GHz), 8 cores (16 threads), 32GB RAM

### DIFF
--- a/calculate_average_gonix.sh
+++ b/calculate_average_gonix.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+
+JAVA_OPTS="--enable-preview"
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Interesting, this one has a huge variance in terms of execution times. Are you observing the same?

gonix:
Interesting. Ignoring the first run with a cold cache (which takes ~4.5s), I used to get variance between ~2.4s to ~2.9s on a consecutive runs, but I was attributing this variance to CPU boost and thermal throttling on my laptop. I assumed thermals would not be a problem on a server, so I do add `sleep 10` between runs to get more stable execution times (idea being that the sleep would allow my laptop CPU to cool down a bit). With the sleep in place I get variance between ~2.4s to ~2.5s with rare dips into 2.3s territory and rare spikes into 2.6s.

Would be really interesting to hear what kind of variance you see on the server side?

gunnarmorling:
It's pretty wild:

```
Benchmark 1: timeout -v 300 ./calculate_average_gonix.sh 2>&1
  Time (mean ± σ):      6.953 s ±  1.144 s    [User: 42.340 s, System: 0.726 s]
  Range (min … max):    5.696 s …  8.092 s    5 runs

Summary
  gonix: trimmed mean 6.993008276126667, raw times 7.12710650246,8.09195019346,5.69645367346,8.00444639346,5.84747193246

Leaderboard

|   | 00:06.993 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_gonix.java)| 21.0.1-open | [gonix](https://github.com/gonix) |  |
```

I also don't think thermal throttling is an issue, I am not observing this for other entries. Anyways, gonna add it. Thx for participating!

gonix:
Super interesting, thanks for sharing!

PR #419, opened 2024-01-15
10k improvement

mtopolnik:
#### Check List:
* Execution time: 3.0
* Execution time of reference implementation: 177

I think the only relevant change is in `nameEquals()`, which now uses `long` loads most of the time.

I delayed submitting this PR since I don't expect much effect on the official dataset, but the 10k dataset should see an improvement. My Hetzner instance suddenly got 2x faster so I'm a bit lost in the effect of this :D

### DIFF
--- a/calculate_average_mtopolnik.sh
+++ b/calculate_average_mtopolnik.sh
@@ -15,7 +15,5 @@
 #  limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:03.376 now, nice improvement!

mtopolnik:
Did you give this code a spin on the 10k dataset? I was really hoping to see an improvement :)

PR #421, opened 2024-01-15
12s (25%) faster on 4 core i7

ianopolous:
* Spread work more evenly across threads
* use ints/shorts instead of longs where possible
* use virtual threads
* remove Map from inner loop - just use single array
* parallelise final merge

Still aiming for simplicity and readability whilst still quite fast. Mostly this means keeping to reading a single byte at a time.

#### Check List:
* Execution time: 36s


gunnarmorling:
00:13.763 now.

PR #424, opened 2024-01-15
jerrinot's initial submission

jerrinot:
#### Check List:
* Execution time: around 5.5s on the ccx33 instance. it's better with just 4 threads, I suspect it's due to hyperthreading which I assume is disabled on the testing machine. Thus this submission uses whatever `Runtime.getRuntime().availableProcessors()` returns.  
* Execution time of reference implementation: loooong :)

This initial submission is aiming to exploit instruction-level parallelism: Each thread pulls from multiple chunks in each iteration. This breaks nice sequential access, but a CPU gets independent streams so it can go bananas with out-of-order processing. 

There are optimization opportunities left on the table, this is to give me some feeling of how it may perform on the testing box. 

Credits:
- @mtopolnik - for many conversations we have had in the last 2 weeks!
- @merykitty - for the amazing branch-less parser
- @royvanrijn - for the hashing-as-you-go idea


gunnarmorling:
Hey @jerrinot, welcome to the party! 00:03.409, nice!

PR #427, opened 2024-01-15
Read file in multiple threads and String to Text

japplis:
#### Check List:
I've run the tests 10 times without errors, let me know the logs if something fails.

* Execution time: 18" old desktop (5" on laptop)
* Execution time of reference implementation: 3'53"

### DIFF
--- a/calculate_average_japplis.sh
+++ b/calculate_average_japplis.sh
@@ -15,5 +15,5 @@
 #  limitations under the License.
 #
 
-JAVA_OPTS="-Xmx2G"
+JAVA_OPTS=""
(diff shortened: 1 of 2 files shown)


gunnarmorling:
This fails validation now, there's one difference reported with the 1B data set used during eval:

```
< Gjoa Haven;-62.7;-14.4;34.3
---
> Gjoa Haven;-62.7;-14.4;39.5
```

Note that there was a rounding bug in the baseline initially which I've fixed (see that new rounding test), caused by inprecise double math. Could it be yours is still subject to this?

japplis:
Thank you for the logs. I think the problem was that combining the stats of each block was not totally thread-safe.
I've adapted the code. Again 10x runs of `tests.sh japplis` + 6x run and compare of 1B rows (using 2 different sets)

PR #433, opened 2024-01-15
jparera's initial implementation

jparera:
#### Check List:
* Execution time: 7s-14s 
* Execution time of reference implementation: 3min 50s
* System specs: 4 core i5-7600k @ 3.8GHz, 32GB RAM

### DIFF
--- a/calculate_average_jparera.sh
+++ b/calculate_average_jparera.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+JAVA_OPTS="--enable-preview --add-modules=jdk.incubator.vector -XX:-TieredCompilation"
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hey, something seems off with the aggregation. If you do _./create_measurements3.sh 1000000000_, do you get the same output as with the baseline (or any of the current leaders for a faster comparison)? I'm seeing differences there.

jparera:
In addition to the rounding problem, the previous code could generate errors and was extremely slow when processing file _measurements3.txt_. The patch should fix all of this. The new code has been tested with _measurements.txt_, _measurements2.txt_, and _measurements3.txt_.

jparera:
I was not satisfied with the code that solves the problem in the output, as it seems to indicate a deeper rounding error problem. So I have delved deeper into its cause.

Let's take the following temperature logs as input:

```
Zürich;18.6
Zürich;10.8
Zürich;11.3
Zurich;-13.6
Zurich;10.5
Zurich;19.7
```

If we do the sum of these `6` records the total is `57.3` which gives an average value of `9.55`. If we round to get a value with a single decimal digit, resolving ties to positive infinity as the `Math.round()` function does, we get `9.6`.

From the [Javadoc of Math.round()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Math.html#round(double)):
> Returns the closest length to the argument, with ties rounding to positive infinity.

I would expect as output:
```
{Zürich=-13.6/9.6/19.7}
```

But the reference code generates:
```
{Zürich=-13.6/9.5/19.7}
```
Surely in this case, the average value was such that there was no tie to resolve. The average value that I get in my new code is `9.549999999999999`, which would be rounded to `9.5` as there is no tie.

gunnarmorling:
Hey, yeah, I think you are right and there's still something off with the rounding behavior of the baseline. I shouldn't have relied on double math 🤷 . FWIW, this specific situation doesn't show up in any of the tests or evaluation files, which is why it hasn't shown up so far. In that light, your implementation looks correct.

gunnarmorling:
00:08.166, nice.

PR #442, opened 2024-01-15
Fix masking

artsiomkorzun:
#### Check List:

* Execution time: same
* Execution time of reference implementation: a lot

One line change, but wasted 2 hours. 

@gunnarmorling should work with this fix.


gunnarmorling:
Nice, can confirm it passes with the 10K keyset again! Time here 00:02.602 i.e. essentially the same as before. Will do another 10K keys run of the top in a bit (hint, hint, @jerrinot ;)), this one may come out really close to the top for that one.

PR #456, opened 2024-01-16
Tonivade

tonivade:
#### Check List:

* Execution time:
```
real	1m53,331s
user	1m47,575s
sys	0m3,467s
```
```
real    3m34,372s
user    3m32,201s
sys     0m3,029s
```
Executed in my laptop:
- Intel Core i7-1065G7
- 32GB RAM
- 4 cores and 2 threads per core

### DIFF
--- a/calculate_average_tonivade.sh
+++ b/calculate_average_tonivade.sh
@@ -15,5 +15,5 @@
 #  limitations under the License.
 #
 
-JAVA_OPTS=""
+JAVA_OPTS="-Xmx1G -Xms1G -XX:+AlwaysPreTouch"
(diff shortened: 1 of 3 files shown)


gunnarmorling:
It seems you're not applying any kind of parallelization? As per the README and PR template, new entries at this point should complete in roughly 10 sec or faster on the eval machine, so as to keep things somewhat manageable. I feel this will run much longer looking at the number you shared. It's not a hard and fast rule, but it should be somewhere up there, which I feel it is not? You can run the current fastest implementation locally to get a feel for the comparison and how long it would run on the eval machine.

tonivade:
Ok, understood! I managed to reduce the time more or less 1 minute, but I doubt that I'm going to reduce it much more and for IMO is likely impossible to reach the 10 seconds. So, I'm going to close the PR.

Thanks a lot for your time, I found it very interesting a I learned a lot of things.

:wave:

PR #457, opened 2024-01-17
test.sh: prepare_<username>.sh should be sourced

plevart:
I don't know whether invoking user's `prepare_<username>.sh` script from the `test.sh` script was meant just to test that the prepare script runs without problems or to actually prepare the user's environment for later running the calculate script. If it was the later, then prepare script should be sourced instead of executed as a sub-process. Only in that case will environment be set for the process (and sub-processes) that are executing the calculate script.

Am I right?

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -15,8 +15,6 @@
 #  limitations under the License.
 #
 
-set -euo pipefail
-
 DEFAULT_INPUT="src/test/resources/samples/*.txt"
 FORK=${1:-""}
 INPUT=${2:-$DEFAULT_INPUT}
@@ -36,8 +34,10 @@ if [ "$#" -eq 0 ] || [ "$#" -gt 2 ] || [ "$FORK" = "-h" ]; then
 fi
 
 if [ -f "./prepare_$FORK.sh" ]; then
-  "./prepare_$FORK.sh"
+  source "./prepare_$FORK.sh"
 fi
+
+set -euo pipefail
 
 for sample in $(ls $INPUT); do
   echo "Validating calculate_average_$FORK.sh -- $sample"


test.sh
```diff
@@ -36,9 +34,11 @@ if [ "$#" -eq 0 ] || [ "$#" -gt 2 ] || [ "$FORK" = "-h" ]; then
 fi
 
 if [ -f "./prepare_$FORK.sh" ]; then
-  "./prepare_$FORK.sh"
+  source "./prepare_$FORK.sh"
 fi
 
+set -euo pipefail
```
gunnarmorling, line 40: Why has that moved?

gunnarmorling:
Ah yes, good spot. It is done correctly in _./evaluate.sh_ (which in turn invokes _./test.sh_), but it should be done as you suggest, so as to support _only_ running _./test.sh_.

ascheman:
I would think that each (relevant) shell script should run autonomously. I guess the workflow most usually is

1. ./prepare... # Sets the Java JDK
2. ./mvnw clean verify # Builds everything based on the set JDK
3. ./test.sh      # Runs all test cases (internally calling prepare again to make sure the JDK is set and calculate... to perform the functional testing)
4. (Optional) Do the performance testing by calling ./calculate...

As step 4 runs outside of test.sh it does not help to set some environment via `source prepare...` during the execution of test.sh. I recommend to make each shell script as autonomous as possible (i.e., set the right env and JDK) and not make them depend on each other.

test.sh
gunnarmorling: @plevart any thought?

gunnarmorling:
@ascheman, I think what you say largely is the case already? Note the key script is _evaluate.sh_ which runs mvn, prepare and test for a given set of entries.

ascheman:
> @ascheman, I think what you say largely is the case already? Note the key script is _evaluate.sh_ which runs mvn, prepare and test for a given set of entries.

Exactly. Therefore, sourcing the prepare script makes no sense. Hence, I recommend to reject the PR.

gunnarmorling:
Ah, gotcha. But in that case we actually should remove the prepare call from test, as it's useless in the current form. That said, I don't think there's any harm with actually do what's proposed here and source the prepare script in test, it just makes for a slight usability improvement, don't you think?

ascheman:
> Ah, gotcha. But in that case we actually should remove the prepare call from test, as it's useless in the current form. That said, I don't think there's any harm with actually do what's proposed here and source the prepare script in test, it just makes for a slight usability improvement, don't you think?

I think, the prepare call should ensure that the test is executed with the right JDK version. So I would leave it here.

I cannot see any usability improvement here? PR #457 does not contain any other change. If someone (for whatever reason) has a reasonable use case, you could re-think the approach. Even then I would rather introduce some optional other script (like `set-test-env.sh` which has the only purpose to set a suitable test environment).

Meanwhile I found that you `source` the prepare file in other places, in particular in [evaluate.sh - line 144](https://github.com/gunnarmorling/1brc/blob/d135bd95214d3c4a2cfa3f0bf70661cee2ebb17f/evaluate.sh#L144) and other places of the evaluate script. On the one hand it would be consequent to source it in test.sh as well to enforce a unique behaviour.

On the other hand, please be aware that sourcing it could do some harm, e.g., defining shell functions to change the behaviour of test.sh or evaluate.sh. Therefore, I would recommend to better change the other places (drop the sourcing of it in `evaluate.sh` and run it as sub shell). Executing it as a sub shell always prevents side effects for the behaviour of the calling script.

Having said that: as long as each change including or changing  prepare-scripts is thoroughly reviewed, the chance to break something is very small.

ascheman:
> > Executing it as a sub shell always prevents side effects for the behaviour of the calling script.
> 
> That's exactly the purpose of the prepare script though: modifying the behavior of the calling script. If it's sub-shell, it wouldn't set the SDK to be used for running the entry later on.

Well, it depends on the surrounding shell. If you start test.sh or evaluate.sh from a shell which is already using sdkman, it doesn't make a difference. Each shell instance which is using sdkman in a regular way already has added directories like $HOME/.sdkman/candidates/java/current/bin to their $PATH. Executing `sdk use java $othercandidate` only switches the symlink `current` to another candidate directory. No environment is set. So you may simply switch (Java) SDK instances in a sub shell and it affects the behaviour in the surrounding environment (parent shell/script), as it only relies on the content of the $PATH variable (which is effectively changed by `sdk use ...` by switching the symbolic link).

This use case is the default for any interactive shell as long as you followed the general sdkman install instructions by sourcing the sdkman init script to your $HOME/.bashrc or $HOME/.bash_profile (or whatever is the right configuration file for your shell).

If you use test.sh or evaluate.sh from a non-interactive shell, you might have to source $HOME/.sdkman/bin/sdkman-init.sh once. But it is still not necessary to source the prepare-script as the relevant additions to $PATH are already performed. Non-interactive shells are used, for example, by cron jobs, executing a remote command via ssh (though for ssh there are other means to initialize the remote shell properly), or GitHub actions.

I just see that for GitHub actions you do exactly this setup in [test_ci.sh](https://github.com/ascheman/1brc/blob/d135bd95214d3c4a2cfa3f0bf70661cee2ebb17f/test_ci.sh#L82-L82). This is the surrounding shell for any other actions. Sourcing the init script sets up $PATH properly. No further sourcing of prepare-scripts necessary beyond this point. You can just run `sdk install ...` as in lines [86](https://github.com/ascheman/1brc/blob/d135bd95214d3c4a2cfa3f0bf70661cee2ebb17f/test_ci.sh#L86-L86) and [94](https://github.com/ascheman/1brc/blob/d135bd95214d3c4a2cfa3f0bf70661cee2ebb17f/test_ci.sh#L94-L94) without any problem. Hence, sourcing the prepare-script as in line [109](https://github.com/ascheman/1brc/blob/d135bd95214d3c4a2cfa3f0bf70661cee2ebb17f/test_ci.sh#L109-L109) is not necessary. You could just execute the prepare-script without any difference.

PR #458, opened 2024-01-17
Try out every tuning under the sun

iziamos:
Hi there, I had some spare time to look at this. It doesn't really constitute a huge code change on my behalf but it may be interesting for you to run as well as it may be the submission with the most flags at this point. Obviously most of these may be silly and won't hold up on the test rig or I hallucinated the improvements.

* I've changed the map size, it helped because I suspect there are fewer hash collisions now.
* I used the number parsing code from the top submission which helped somewhat. There was no point trying to write that myself as it seems to be a solved problem now.
* I've added every tuning flag that seemed to affect execution time on my computer.

#### Check List:

* Execution time: ranges between 1-1.1 sec on my system.
* Execution time of reference implementation: 47 sec.

I'd predict under 5 seconds for this one given the new parsing code but it would be interesting to see how far the tuning gets it.

Also if it's not too much of an ask I think it would be interesting to see how long the JVM with --dry-run enabled takes on the test rig. I see:

real	        0m0.019s
user	0m0.012s
sys	        0m0.011s

locally. Does this mean the jvm takes that long to load or does it just exit after parsing the arguments? One would have to take a look but  simple hello world does it in iirc 300 millis on my system.

Would be cool if you could get those numbers, totally understand if you are swamped by this whole thing though.

Thanks

### DIFF
--- a/calculate_average_iziamos.sh
+++ b/calculate_average_iziamos.sh
@@ -15,5 +15,11 @@
 #  limitations under the License.
 #
 
-JAVA_OPTS="--enable-preview --add-modules=jdk.incubator.vector -Djdk.incubator.vector.VECTOR_ACCESS_OOB_CHECK=0 -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -Xms16m -Xmx16m -XX:-AlwaysPreTouch -XX:-TieredCompilation -XX:CICompilerCount=1"
+JAVA_OPTS="--enable-preview
+  -XX:+UnlockExperimentalVMOptions \
+  -XX:+UseEpsilonGC -Xms16m -Xmx16m -XX:-AlwaysPreTouch \
+  -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:CompilationMode=high-only \
+  -XX:C1MaxTrivialSize=500 -XX:-UseCountedLoopSafepoints -XX:+UseCMoveUnconditionally -XX:+DisableAttachMechanism \
+  -XX:-PreserveFramePointer -Xnoclassgc -disablesystemassertions -XX:-UsePerfData  \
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:04.154 now, very nice.

Not sure what you mean by --dry-run, but a simple --version yields this:

```
time java --version
openjdk 21.0.1 2023-10-17
OpenJDK Runtime Environment (build 21.0.1+12-29)
OpenJDK 64-Bit Server VM (build 21.0.1+12-29, mixed mode, sharing)

real	0m0.032s
user	0m0.020s
sys	0m0.016s
```

iziamos:
Hi @gunnarmorling sorry for bumping but thanks for running this version!. I just discovered dry run the other day while going through all the flags. I was asking cause I suppose its interesting to see how much of the fastest solution is VM startup. The description goes as such:

--dry-run
    Creates the VM but doesn't execute the main method. This --dry-run option might be useful for validating the command-line options such as the module system configuration. 

On the machine I'm on at the moment, for dry run I get:

real	 0m0.036s
user 0m0.024s
sys	 0m0.021s

vs 

[ziamosj@work ~]$ time java --version
openjdk 21.0.1 2023-10-17 LTS
OpenJDK Runtime Environment Zulu21.30+15-CA (build 21.0.1+12-LTS)
OpenJDK 64-Bit Server VM Zulu21.30+15-CA (build 21.0.1+12-LTS, mixed mode, sharing)

real	   0m0.028s
user   0m0.008s
sys	   0m0.022s

So I suspect it does abit more work than version. 

This is going in the long list of things I've learned as part of this :D

gunnarmorling:
Oh, interesting, didn't know that either. Thanks, TIL!

PR #462, opened 2024-01-17
A fast implementation without unsafe

ianopolous:
I wanted to preserve my existing simple implementation, hence the 'fast' suffix. (Github user ianopolousfast doesn't exist)

Hash collisions are handled on line 93.

#### Check List:

* Execution time: 19s on 4 core i7

### DIFF
--- a/calculate_average_ianopolousfast.sh
+++ b/calculate_average_ianopolousfast.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:08.045. Seems I should have put a "one entry per person" rule in place, but I haven't, so I'm gonna merge it ;)

ianopolous:
Thank you for all you're hard work! It's been great fun and I've learnt a lot!

My biggest takeaway for the JVM is that Java should allow memory mapping a file bigger than 2 GiB.

gunnarmorling:
> Thank you for all your hard work! It's been great fun and I've learnt a lot!

Awesome, that's great to hear!

> My biggest takeaway for the JVM is that Java should allow memory mapping a file bigger than 2 GiB.

Ha, in fact it's possible now with the new incubating foreign memory API (https://openjdk.org/jeps/442).

PR #471, opened 2024-01-18
fix test rounding, pass 10K station names

jgrateron:
Minor fixes and the execution time is almost the same, 

I only corrected the rounding test and that it works with the 10,000 station names without errors of overflow memory heap.


gunnarmorling:
00:17.179, nice!

PR #472, opened 2024-01-18
using unsafe,graal native etc

karthikeyan97:
changing using mappedbytebuffer

changes before using unsafe address

using unsafe

#### Check List:

* Execution time:1273ms
* Execution time of reference implementation:208000ms


prepare_karthikeyan97.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
ascheman, line 1: The file (currently) needs execute permissions, otherwise the `test.sh` script will not use it (-> @gunnarmorling, this could be enhanced in test.sh as well).

calculate_average_karthikeyan97.sh
```diff
@@ -16,4 +16,14 @@
 #
 
 JAVA_OPTS="-Xms20480m -Xmx40960m "
-java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_karthikeyan97
+
+if [ -f target/CalculateAverage_karthikeyan97_image ]; then
+    #echo "Picking up existing native image 'target/CalculateAverage_karthikeyan97_image', delete the file to select JVM mode." 1>&2
+    target/CalculateAverage_karthikeyan97_image -Xms20480m -Xmx40960m
```
ascheman, line 22: To my understanding, the max. memory is restricted to 32GB, @gunnarmorling? Hence, requesting 40GB of Heap does not work.

ascheman (changes requested):
Please have a look for my comments, @gunnarmorling

calculate_average_karthikeyan97.sh
```diff
@@ -16,4 +16,14 @@
...
+else
+    JAVA_OPTS="--enable-preview"
+    #echo "Chosing to run the app in JVM mode as no native image was found, use prepare_karthikeyan97.sh to generate." 1>&2
+    java -Xms20480m -Xmx40960m --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_karthikeyan97
```
ascheman, line 26: Running with JVM aborts with 

java.lang.UnsupportedClassVersionError: Preview features are not enabled for dev/morling/onebrc/CalculateAverage_karthikeyan97 (class file version 65.65535). Try running with '--enable-preview'

prepare_karthikeyan97.sh
ascheman: BTW: Executing test.sh reveals an error
```
Validating calculate_average_karthikeyan97.sh -- src/test/resources/samples/measurements-1.txt
1c1
< 
---
> Kunming;19.8;19.8;19.8
```

calculate_average_karthikeyan97.sh
gunnarmorling: That's the max heap size, so no guarantee that the app will get that much, depending on the system. The environment has 128G, so it actually may get the full 40 (of course it's an excessively large value).

calculate_average_karthikeyan97.sh
gunnarmorling: Yeah, I thought I had pointed that out before, but somehow my comment got lost.

@karthikeyan97 please make sure that _./test.sh_ passes without differences and that the output for 1B rows file matches that one of the baseline (or those currently fastest entries to be a bit quicker).

karthikeyan97:
Sorry for not correcting this.Will fix this

PR #477, opened 2024-01-18
Feature/review made easy

ascheman:
These scripts support local code reviews (cf. contained README).

They are build upon other changes to facilitate tests with (large) measurement files (>= 100.000 entries) with unique measurements. The idea behind it was to detect hash function/collision errors which are more probable to occur with larger unique data sets.

But it should be possible to split these two concerns, @gunnarmorling or at least to make the execution of the _unique_ tests optional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,7 @@ out/
 /*.out
 out_expected.txt
 /*-timing.json
+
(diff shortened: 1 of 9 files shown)


ascheman:
Withdrawn

PR #478, opened 2024-01-18
3s (16%) faster, still no unsafe

ianopolous:
* use Arena and MemorySegment to map entire file at once
* handle station names < 8 and < 16 separately
* attempt branchless temperature parsing
* store first two longs of name for faster comparison
* switch to normal threads and 1 task per thread

Hash collisions handled at lines 115, 137 and 159

Thanks for the pointer to https://openjdk.org/jeps/442 @gunnarmorling !

#### Check List:

* Execution time: 16s on 4core i7

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_ianopolousfast.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_ianopolousfast.java
@@ -15,45 +15,53 @@
  */
 package dev.morling.onebrc;
 
-import java.io.*;
-import java.nio.*;
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.nio.ByteOrder;
 import java.nio.channels.*;
-import java.util.concurrent.*;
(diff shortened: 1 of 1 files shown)


gunnarmorling:
It's a tad faster now (00:13.071) on the eval machine, not quite the big difference you've observed locally.

> Thanks for the pointer to https://openjdk.org/jeps/442 @gunnarmorling !

👍

ianopolous:
I suspect you ran my original entry without the fast suffix? The previous run with fast was 8s.
So ./calculate_average_ianopolousfast.sh

gunnarmorling:
Whoops, that's what I did indeed. Good catch. The "fast" entry comes in at 00:07.202 now. Still not 16%, but definitely a significant improvement. If you'd like to do more updates (definitely welcome to), I'd suggest we drop your non-fast one, as it's a bit in the way of how the eval tooling works, requiring some more manual adjustment to the leaderboard snippet.

PR #492, opened 2024-01-19
Plain IO

artsiomkorzun:
#### Check List:

* Execution time: shows a good gain locally
* Execution time of reference implementation: a lot

@gunnarmorling I hope this time I did not add any bugs:)

### DIFF
--- a/calculate_average_artsiomkorzun.sh
+++ b/calculate_average_artsiomkorzun.sh
@@ -17,9 +17,9 @@
 
 if [ -f target/CalculateAverage_artsiomkorzun_image ]; then
     echo "Picking up existing native image 'target/CalculateAverage_artsiomkorzun_image', delete the file to select JVM mode." 1>&2
-    target/CalculateAverage_artsiomkorzun_image
+    target/CalculateAverage_artsiomkorzun_image -XX:MaxDirectMemorySize=4294967296
 else
-    JAVA_OPTS="--enable-preview -Xmx128m -XX:+UseSerialGC -XX:-TieredCompilation"
+    JAVA_OPTS="--enable-preview -Xmx128m -XX:+UseSerialGC -XX:-TieredCompilation -XX:MaxDirectMemorySize=4294967296"
     echo "Chosing to run the app in JVM mode as no native image was found, use prepare_artsiomkorzun.sh to generate." 1>&2
     java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_artsiomkorzun
 fi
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Yepp, that looks good, really good :)

```
Benchmark 1: timeout -v 300 ./calculate_average_artsiomkorzun.sh 2>&1
  Time (mean ± σ):      2.463 s ±  0.008 s    [User: 17.837 s, System: 1.697 s]
  Range (min … max):    2.454 s …  2.484 s    10 runs

Summary
  artsiomkorzun: trimmed mean 2.4611661857500002, raw times 2.464730524,2.458767629,2.46014498,2.46151103,2.462863934,2.484195729,2.453602039,2.461667844,2.460033884,2.459609661

Leaderboard

|   | 00:02.461 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_artsiomkorzun.java)| 21.0.1-graal | [Artsiom Korzun](https://github.com/artsiomkorzun) | GraalVM native binary |
```

franz1981:
Hi @artsiomkorzun I see you have used aligned buffers, but I am fairly surprised that it is faster than mmap here? Can you share the motivation why you choose it?
In theory NIO still have to perform a cheap copy Vs the OS page cache, and using a syacall, too (I didn't verified, but it should), while mmap, especially against tmpfs *is* the OS page cache ie no copy, no syscalls.

artsiomkorzun:
@franz1981 the benchmark script tests the process execution time, not when the output appears in std out. So when a process finishes the mapped pages get unmapped and it takes quite a bit of time (200 - 400 ms on my machine). When you try to unmap pages by hand, you will find out that munmap takes "some global process lock", so it is not scalable at all (you can't parallelize it). So you have two options:

1. To start a subprocess to hide munmap cost. (Not sure if this allowed)
2. To read these pages and copy them in user space, which is scalable and runs faster than munmap with this number of threads.

gunnarmorling:
Ah, that's very interesting. Turns out @thomaswue is exploring route 1 ("To start a subprocess to hide munmap cost. (Not sure if this allowed)") [right now](https://twitter.com/thomaswue/status/1748787118560686190).

artsiomkorzun:
@gunnarmorling seems like you will be busy on Monday to merge a lot of pr's with this subprocess thing=)

gunnarmorling:
Hehe, yeah, that's certainly a concern I have... ;)

franz1981:
Thanks @artsiomkorzun but, what about mmap a single big chunk via segment and forcibly kill the process causing a SIGSEV accessing some bad address before the main exit kick-in? This should avoid the munmap to happen no?

artsiomkorzun:
@franz1981 OS cleans up resources after a process finishes, including mapped pages. This time is included in process execution time.

franz1981:
Thanks for the clarification @artsiomkorzun !!

PR #499, opened 2024-01-20
yonatang solution: a jdk8 friendly, no unsafe code, epsilon-gc friendly solution

yonatang:
#### Check List:
- [X] Your launch script is named `calculate_average_yonatang.sh` (make sure to match casing of your GH user name) and is executable
- [X] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number) - [line 61](https://github.com/gunnarmorling/1brc/blob/5babc05182e7687832aada9cf57157b1e9ab3992/src/main/java/dev/morling/onebrc/CalculateAverage_yonatang.java#L61), the `station` string is linearly transformed to a `long[13]`. if there is a hash collision, it will jump to the next empty space.

* Execution time: **27.16 seconds** (MacBook pro 2019 i9) - about ~25% slower then one of the faster solutions, probably around 4s on the evaluation machine.
* Execution time of reference implementation: Merykitty's - **22.04 seconds** (the safe, openjdk version)

### A few words

I took the challenge of writing a "corporate friendly" solution: something that will run reasonably fast, without using any unsafe code, that can be compiled and executed with JDK 8. Just trying to make a point that corporate code can still be fast, if you really want to.

In order to compensate these limitations, I implemented an allocation-free solution (almost everything is on the stack), that could leverage the Epsilon GC, when running with a modern JVM. Using Epsilon GC reduces processing time by ~20%, comparing to G1GC. I believe the dominant contributing factor is not directly the GC overhead reduction, but it is the elimination of the CPU context switching, which can invalidate the CPU L1/L2 caching.


gunnarmorling:
Thx for participating and explaining the reasoning! Loving to see the usage of EpsilonGC. This comes in at  00:12.800, i.e. a bit slower (relatively speaking) than what you have observed. Could be due to differences in ISA, e.g. longer vector registers on the eval machine from which the top entries benefit disproportionally more.

PR #502, opened 2024-01-20
Improved integer conversion, delayed string creation.

jgrateron:
All test pass
new runtime 14 sec.


gunnarmorling:
Interesting, is this actually faster on your system? On the eval system it's ~19 sec now.

jgrateron:
start my linux in "init 3" defaults the results, now if the time is correct 14 seconds

gunnarmorling:
How is this dealing with hash collisions actually? It seems if two station names yield the same hash code, the aggregation would be wrong?

jgrateron:
I did several tests and there are no collisions between station names, but to prevent this I added the length of the string in the initialization of the hash.

image

gunnarmorling:
Hum, I fail to understand how this resolves the issue. As a matter of principle it must be guaranteed that no values from several stations are mixed up, i.e. all potential collisions must be handled correctly.

PR #503, opened 2024-01-20
Add linl33's implementation

linl33:
#### Check List:

Linear probing loop  
https://github.com/gunnarmorling/1brc/blob/bbf7537d6859a54c3d3e0aea12933721b0f5dc1b/src/main/java-22/dev/morling/onebrc/CalculateAverage_linl33.java#L314

Collision check  
https://github.com/gunnarmorling/1brc/blob/bbf7537d6859a54c3d3e0aea12933721b0f5dc1b/src/main/java-22/dev/morling/onebrc/CalculateAverage_linl33.java#L322

String comparison  
https://github.com/gunnarmorling/1brc/blob/bbf7537d6859a54c3d3e0aea12933721b0f5dc1b/src/main/java-22/dev/morling/onebrc/CalculateAverage_linl33.java#L390

* Execution time:
```
Benchmark 1: ./calculate_average_linl33.sh
  Time (mean ± σ):      3.652 s ±  0.174 s    [User: 16.066 s, System: 1.022 s]
  Range (min … max):    3.295 s …  3.901 s    20 runs
```
```
 Performance counter stats for './calculate_average_linl33.sh':

         16,291.13 msec task-clock                       #    4.805 CPUs utilized             
             8,473      context-switches                 #  520.099 /sec                      
               227      cpu-migrations                   #   13.934 /sec                      
           261,538      page-faults                      #   16.054 K/sec                     
    52,471,272,103      cycles                           #    3.221 GHz                       
   104,975,725,889      instructions                     #    2.00  insn per cycle            
    14,246,140,162      branches                         #  874.472 M/sec                     
       586,828,252      branch-misses                    #    4.12% of all branches           

       3.390802376 seconds time elapsed

      15.212102000 seconds user
       1.071132000 seconds sys
```
```
Benchmark 1: ./calculate_average_merykittyunsafe.sh
  Time (mean ± σ):      3.751 s ±  0.174 s    [User: 16.371 s, System: 1.019 s]
  Range (min … max):    3.338 s …  4.102 s    20 runs
```

I put my Java class under `src/main/resources/` because it requires jdk 22 to compile. `prepare_linl33.sh` handles compiling the class with jdk 22.

Please take a look, thanks for organizing this competition!

### DIFF
--- a/evaluate.sh
+++ b/evaluate.sh
@@ -36,7 +36,7 @@ RESET='\033[0m' # No Color
 MEASUREMENTS_FILE="measurements_1B.txt"
 RUNS=5
 DEFAULT_JAVA_VERSION="21.0.1-open"
-BUILD_JAVA_VERSION="21.0.1-open"
+: "${BUILD_JAVA_VERSION:=21.0.1-open}"
 RUN_TIME_LIMIT=300 # seconds
 
 TIMEOUT=""
--- a/evaluate.sh
+++ b/evaluate.sh
@@ -36,6 +36,7 @@ RESET='\033[0m' # No Color
 MEASUREMENTS_FILE="measurements_1B.txt"
 RUNS=5
 DEFAULT_JAVA_VERSION="21.0.1-open"
+BUILD_JAVA_VERSION="21.0.1-open"
 RUN_TIME_LIMIT=300 # seconds
 
 TIMEOUT=""
@@ -115,6 +116,7 @@ if [ -f "/sys/devices/system/cpu/cpufreq/boost" ]; then
   fi
 fi
 
+print_and_execute sdk use java $BUILD_JAVA_VERSION
 print_and_execute java --version
 print_and_execute ./mvnw --quiet clean verify
 
--- a/prepare_linl33.sh
+++ b/prepare_linl33.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+
+source "$HOME/.sdkman/bin/sdkman-init.sh"
+# TODO: bump to ea 32 when available
+sdk use java 22.ea.31-open 1>&2
+
+CLASS_NAME="CalculateAverage_linl33"
+
+JAVA_OPTS="-Xrs --enable-preview --add-modules jdk.incubator.vector --enable-native-access=ALL-UNNAMED"
+JAVA_OPTS="${JAVA_OPTS} -XX:+UnlockDiagnosticVMOptions -XX:+UnlockExperimentalVMOptions"
+JAVA_OPTS="${JAVA_OPTS} -Xms128m -XX:+AlwaysPreTouch -XX:+AlwaysPreTouchStacks -XX:-UseTransparentHugePages"
+JAVA_OPTS="${JAVA_OPTS} -XX:-UseCompressedClassPointers -XX:+ForceUnreachable -XX:-CompactStrings"
+JAVA_OPTS="${JAVA_OPTS} -XX:CodeEntryAlignment=64 -XX:OptoLoopAlignment=64 -XX:MaxLoopPad=16 -XX:ObjectAlignmentInBytes=64"
+JAVA_OPTS="${JAVA_OPTS} -XX:-UseLoopPredicate -XX:LoopStripMiningIter=0 -XX:LoopStripMiningIterShortLoop=0"
+JAVA_OPTS="${JAVA_OPTS} -XX:-UseCountedLoopSafepoints -XX:GuaranteedSafepointInterval=0 -XX:AllocatePrefetchStyle=0"
+JAVA_OPTS="${JAVA_OPTS} -XX:+TrustFinalNonStaticFields -XX:LockingMode=2 -XX:+UseSystemMemoryBarrier"
+JAVA_OPTS="${JAVA_OPTS} -XX:-UseDynamicNumberOfCompilerThreads -XX:-UseDynamicNumberOfGCThreads"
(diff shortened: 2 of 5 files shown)


gunnarmorling:
Hey, thx a lot for this entry! What is the specific reason for requiring Java 22? We can do it if necessary, but it'll require a bit of adjustment.

Specifically, I think the following should be done:

* Make the Java version in the _pom.xml_ file overridable by means of a build property (still defaulting to Java 21)
* Add a new source folder _src/main/java-22/_ and move your implementation there
* Add a Maven build profile which adds this source folder when the build property is set to 22
* Add a constant at the top of _evaluate.sh_ for the Java version with value of 21, and pass that version to the `mvnw` invocation in that script

That way, it should be possible to keep everything as is for Java 21, but entries like yours could be run with 22 by adjusting the constant in _evaluate.sh_.

Could you rework your PR along those lines? Thx!

linl33:
I'm using Java 22 for the Panama API. There are changes made in 22 that I find makes it a bit easier to work with.

I will rework the PR like you outlined.

The reason why I put the build steps in `prepare_linl33.sh` (instead of letting Maven build) is so that when running `evaluate.sh`, the Java version doesn't need to be manually switched to 22 beforehand. When the prepare script is sourced, it can automatically switch to 22, compile, then update the JAR file with the new classes.

gunnarmorling:
> I'm using Java 22 for the Panama API. There are changes made in 22 that I find makes it a bit easier to work with.
>
> I will rework the PR like you outlined.

Appreciate it, thx!

linl33:
@gunnarmorling I've added the following
 - A Maven profile `jdk22` that activates when Java 22 is used to run Maven (or it can be explicitly activated with `-Pjdk22`)
 - A new parameter `BUILD_JAVA_VERSION` in `evaluate.sh`, which gets activated via `sdk use` before executing Maven.

prepare_linl33.sh
```diff
@@ -0,0 +1,45 @@
...
+JAVA_OPTS="${JAVA_OPTS} -XX:-UseDynamicNumberOfCompilerThreads -XX:-UseDynamicNumberOfGCThreads"
+JAVA_OPTS="${JAVA_OPTS} -XX:ArchiveRelocationMode=0 -XX:-UsePerfData -XX:-UseNotificationThread -XX:-CheckIntrinsics"
+#JAVA_OPTS="${JAVA_OPTS} -XX:+UseZGC -XX:-ZProactive -XX:+ZCollectionIntervalOnly -XX:ZCollectionInterval=0 -XX:-ZUncommit -XX:-ZBufferStoreBarriers -XX:ZIndexDistributorStrategy=1"
+JAVA_OPTS="${JAVA_OPTS} -XX:+UseEpsilonGC -XX:-UseCompressedOops"
+#JAVA_OPTS="${JAVA_OPTS} -XX:+UseParallelGC -XX:-UseCompressedOops"
+#JAVA_OPTS="${JAVA_OPTS} -XX:+UseG1GC -XX:-UseCompressedOops"
+JAVA_OPTS="${JAVA_OPTS} -Djdk.incubator.vector.VECTOR_ACCESS_OOB_CHECK=0 -Djava.lang.invoke.VarHandle.VAR_HANDLE_GUARDS=false -Djava.lang.invoke.MethodHandle.DONT_INLINE_THRESHOLD=-1"
+JAVA_OPTS="${JAVA_OPTS} -Dfile.encoding=UTF-8 -Dsun.stdout.encoding=UTF-8 -Dsun.stderr.encoding=UTF-8"
+
+# create CDS archive
```
gunnarmorling, line 42: How important is CDS to you? It's a bit of a problem right now when running _./evaluate.sh linl33_, as the test script removes the linked measurements file and thus evaluate.sh fails.

evaluate.sh
```diff
@@ -89,7 +90,12 @@ if [ ! -d "$HOME/.sdkman/candidates/java/$DEFAULT_JAVA_VERSION" ]; then
   print_and_execute sdk install java $DEFAULT_JAVA_VERSION
 fi
 
-# 4. Install missing SDK java versions in any of the prepare_*.sh scripts for the provided forks
+# 4. make sure the build java version is installed
+if [ ! -d "$HOME/.sdkman/candidates/java/$BUILD_JAVA_VERSION" ]; then
+  print_and_execute sdk install java $BUILD_JAVA_VERSION
+fi
```
gunnarmorling, line 96: That should not be needed, as all JDKs referenced by all prepare scripts will be installed already.

prepare_linl33.sh
linl33: On my laptop the difference is about 200ms.

I didn't realize measurement.txt might not exist seeing that prepare runs before test. I will change it to read from one of the test files in resources when building the CDS archive.

evaluate.sh
linl33: 👍 will remove it

prepare_linl33.sh
linl33: Fixed in the latest push

evaluate.sh
linl33: Removed in the latest push

evaluate.sh
```diff
@@ -36,6 +36,7 @@ RESET='\033[0m' # No Color
 MEASUREMENTS_FILE="measurements_1B.txt"
 RUNS=5
 DEFAULT_JAVA_VERSION="21.0.1-open"
+BUILD_JAVA_VERSION="21.0.1-open"
```
gunnarmorling, line 39:
```suggestion
: "${BUILD_JAVA_VERSION:=21.0.1-open}"
```

That way, it can be overridden when launching:

```
BUILD_JAVA_VERSION=22.ea.31-open ./evaluate.sh linl33
```

prepare_linl33.sh
gunnarmorling: Perfect, thx!

gunnarmorling:
00:03.013, very nice! To evaluate any entry like this which requires another Java version for building, it can now be run like so:

```
BUILD_JAVA_VERSION=22.ea.31-open ./evaluate.sh linl33
```

Out of curiosity, which changes from Java 22 did you require?

linl33:
@gunnarmorling Could you put my entry on the leaderboard when you update the leaderboard?

gunnarmorling:
Oh, apologies for that, not sure why it went missing. Just added it. Thx for the call-out!

linl33:
Thanks!

tivrfoa:
> Oh, apologies for that, not sure why it went missing. Just added it. Thx for the call-out!

Maybe it's because his source is in `src/main/java-22` , and the script might be not looking there.

gunnarmorling:
No, it can't be that. The leaderboard output is generated by evaluate.sh,
and it definitely is there for this entry. I must have just forgotten it :)

PR #507, opened 2024-01-20
Solution without unsafe

giovannicuccu:
#### Check List:

Dealing with  hash collisions starts at line 195

* Execution time: 5.5 secs
* Execution time of reference implementation:143 secs

Execution times recorded using Windows

My execution time is comparable to the spullara's one so I expect it running in less than 10 seconds.


gunnarmorling:
00:07.240, nice!

PR #509, opened 2024-01-20
reorganize code, little bit faster

abeobk:
#### Check List:

Standard set
Benchmark 1: ./calculate_average_abeobk.sh
  Time (mean ± σ):     736.2 ms ±  17.3 ms    [User: 18932.3 ms, System: 722.9 ms]
  Range (min … max):   716.6 ms … 767.5 ms    10 runs

10k set
Benchmark 1: ./calculate_average_abeobk.sh
  Time (mean ± σ):      1.848 s ±  0.022 s    [User: 49.804 s, System: 0.985 s]
  Range (min … max):    1.816 s …  1.889 s    10 runs

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_abeobk.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_abeobk.java
@@ -47,6 +47,10 @@ public class CalculateAverage_abeobk {
             0xffffffffffffffL,
             0xffffffffffffffffL, };
 
+    private static final void debug(String s, Object... args) {
+        System.out.println(String.format(s, args));
+    }
+
     private static Unsafe initUnsafe() {
         try {
             Field theUnsafe = Unsafe.class.getDeclaredField("theUnsafe");
@@ -110,7 +114,7 @@ public class CalculateAverage_abeobk {
         }
 
         boolean contentEquals(long other_addr, long other_tail) {
-            if (tail != other_tail) // compare tail & length at the same time
+            if (tail != other_tail)
                 return false;
             // this is faster than comparision if key is short
             long xsum = 0;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_abeobk.java
```diff
@@ -330,13 +330,16 @@ public static void main(String[] args) throws InterruptedException, IOException
             }
 
             if (SHOW_ANALYSIS) {
-                System.out.println("total=" + Arrays.stream(lenhist).sum());
-                System.out.println("length_histogram = "
+                debug("Collision stat: ");
```
gunnarmorling, line 333: Just to be sure: does your implementation guarantee to handle collisions 100% correctly? Or can there be a (however small) risk for unhandled collisions and thus incorrect results? I suppose it's the former.

abeobk: Yes it is handled correctly (with the assumptions given in the 1brc description).
Since there are only10k stations at most, the hash table size is intentionally added 10000, so there is no need to mod the table size.  The performance gains was achieved by using a fast/low collision mixer ( 0 collision for standard set, still low for 10k set, I had some fun tweaking this function :) )  and by removing lots of unnecessary branching.

gunnarmorling: Got it, thx.

gunnarmorling:
Nice, 00:02.336 now!

```
Benchmark 1: timeout -v 300 ./calculate_average_abeobk.sh 2>&1
  Time (mean ± σ):      2.336 s ±  0.006 s    [User: 15.820 s, System: 0.701 s]
  Range (min … max):    2.329 s …  2.343 s    5 runs

Summary
  abeobk: trimmed mean 2.3362541187466674, raw times 2.3407769550800004,2.32911485408,2.3351161250800003,2.3429942460800004,2.3328692760800003

Leaderboard

|   | 00:02.336 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_abeobk.java)| 21.0.1-graal | [Van Phu DO](https://github.com/abeobk) | GraalVM native binary |
```

PR #517, opened 2024-01-20
Read and parsing same task, RandomAccessFile and ByteBuffer

japplis:
- Read and parsing block done in the same task
- Use RandomAccessFile instead of FileInputStream
- Use ByteBuffer instead of Text (7" on laptop)
 
#### Check List:
ByteBuffer is now used as for hashcode and equals

* Execution time: 7"
* Execution time of reference implementation: 2"43

Execution time is a bit slower but hoping it will be better on the server as I see some differences between implementations, sometimes twice slower, sometimes twice faster.


gunnarmorling:
Definitely slower:  00:15.317. I reckon you don't want to add this as-is?

japplis:
Yes, indeed. Let's not merge it if it's slower.

PR #525, opened 2024-01-21
Reduce allocations and heap size

roman-r-m:
#### Check List:

Minus another 100-150ms on my machine.

* Execution time:

### DIFF
--- a/calculate_average_roman-r-m.sh
+++ b/calculate_average_roman-r-m.sh
@@ -19,6 +19,6 @@ JAVA_OPTS="--enable-preview -XX:+UseTransparentHugePages"
 
 # epsilon GC needs enough memory or it makes things worse
 # see https://stackoverflow.com/questions/58087596/why-are-repeated-memory-allocations-observed-to-be-slower-using-epsilon-vs-g1
-JAVA_OPTS="$JAVA_OPTS -XX:+UnlockExperimentalVMOptions -XX:-EnableJVMCI -XX:+UseEpsilonGC -Xmx64M -Xms64M -XX:+AlwaysPreTouch"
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hey, this one produces an incorrect output for the 10K keyset test (see create_measurements3.sh). While that's not the official challenge, I'd like to make sure that at least the top entries pass that, so as to make sure they don't cut any corners. This one runs out of heap space for that test, so it might bea simple fix. Could you take a look? Thx!

roman-r-m:
Hmm, seems to be working for me.
Just to confirm, it's `measurements3.txt` right? Maybe there's some difference in how I run it. Is there a script to do it or is it all manual?

roman-r-m:
I don't want to just increase the heap size more than needed, because doing so makes the result of the small dataset worse. Not sure why.

gunnarmorling:
> Just to confirm, it's measurements3.txt right?

No, I was referring to a 1B rows file with 10K keys. You can create one via ./create_measurements3.sh 1000000000. This PR OOME's for that right now.

roman-r-m:
> No, I was referring to a 1B rows file with 10K keys. You can create one via ./create_measurements3.sh 1000000000. This PR OOME's for that right now.

I may be missing something but `create_measurements3.sh` calls `CreateMeasurements3` which writes to `measurements3.txt`:

```java
try (var out = new BufferedWriter(new FileWriter("measurements3.txt"))) {
```

roman-r-m:
Anyway, I've increased the heap size, could you try again pls.
I'll try to replicate it locally later today.

gunnarmorling:
Ah, sorry, I thought you were referring to _src/test/resources/samples/measurements-3.txt_ (which is just one of the test cases). So yes, that measurements3.txt file created by _create_measurements3.sh_ was failing for me (after soft-linking it to measurements.txt).

Will run again as per your latest update.

gunnarmorling:
00:04.551 now, also passing the 10K test (in ~20 sec).

PR #527, opened 2024-01-21
Added implementation CalculateAverage_kaufco - The Bit Bang Theory

kaufco:
Bits are heavily bashed, twiddled, smashed, banged and jammed. Twisted, crunched, flipped and squashed. This implementation tries some serious bit trickery and scores an execution time of 2.05s (**update: now 1.32s**) on my machine. It does not resort to vector API, Unsafe or GraalVM native binaries. Just Good ol' OpenJDK.

#### Key ideas of this implementation:
- Do not scan the strings bytewise in a loop. There is a function `scanFast` that scans 8 bytes at once.
- Fixed point integer calculation instead of floating point. Biased values to represent negatives.
- Quirky number formats such as 11 bit integers. Bit fields packed into 64 bit words to have fewer transfers to the data buffers. Bit fiddling costs nothing on a CPU. Local vars are mapped to registers. Buffer transfers is what is expensive, especially when you don't use Unsafe :-)

#### Check List:

#### Benchmark:
* Execution time: 2.05s (**update: 1.32s after last commit**)
  * baseline: 2m58.42s
  * merykitty: 2.11s
  * merykittyunsafe: 1.83s

#### My test environment:
* AMD Ryzen 9 3950x (16 Cores)
* 32GB RAM
* an SSD (no RAMdisk used, but after first run the file is in the file system cache)
* Ubuntu 23.04
* OpenJDK Runtime Environment Corretto-21.0.1.12.1


tejas-2232:
Great test environment. 💯 
This is not related to the problem, but I would like to ask you some questions. 
How long you've been using Ubuntu 23.04? Is it smooth? what's your review?

kaufco:
> Great test environment. 💯 This is not related to the problem, but I would like to ask you some questions. How long you've been using Ubuntu 23.04? Is it smooth? what's your review?

Thanks :-) I just upgraded to 23.04 last week and don't see much difference to the previous version. I'm using Ubuntu since 2009 though. It handles the many cores very smoothly, but I cannot tell really whether that would not be the case for other OS.

prepare_kaufco.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
gunnarmorling, line 1: No need for this if it's actually empty.

calculate_average_kaufco.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Capture the PID of the Java process
+JAVA_PID=$!
+
+# Monitor the output file for the desired character
+tail -f java_output.log | {
+    read -r line
+    echo "$line"
+    kill -9 $JAVA_PID
+}
```
gunnarmorling, line 29: What's the purpose of this, can it be removed?

src/main/java/dev/morling/onebrc/CalculateAverage_kaufco.java
```diff
@@ -0,0 +1,643 @@
...
+                                     ByteBuffer stationNames,
+                                     int hash,
+                                     int temperature) {
+        int ofs = (hash & WEATHER_DATA_INDEX_MASK) << 4;
+        long lo = weatherData.getLong(ofs);
+        long hi = weatherData.getLong(ofs + 8);
+
+        int prevHash = (int) (lo >> 32);
+        if ((prevHash != hash) && (prevHash != 0))
+            throw new HashCollisionException();
```
gunnarmorling, line 375: So what happens if there is a collision? Is there some means of open addressing fallback or similar? Can you point me to it? Note that the program should work for any valid keyset, so collisions must be handled somehow. To that end, when I test with 10K keys (see _create_measurements3.sh_), the program doesn't seem to finish (or it takes a very long time). While that's not the official evaluation, I'd like to make sure at least for the top entries that this keyset also can be processed, so as to make sure no corners are cut. Could you take a look? Thx.

gunnarmorling (commented):
A few comments inline.

src/main/java/dev/morling/onebrc/CalculateAverage_kaufco.java
kaufco: Yes: it falls back to the baseline implementation (`CalculateAverage_baseline`), so it's normal that it is getting slow. These are the relevant lines:

692 to 609:
```
        public void run() {
            try {
                processFile();
            }
            catch (IOException | HashCollisionException e) {
                abortRun = true;
            }
        }
```

and 202..206:
```
        if (abortRun) {
            // In case of hash collision: slooow fallback
            CalculateAverage_baseline.main(new String[0]);
            return;
        }
```

My original intention was to resort to a multi-set hash table implementation in the fallback case (which would probably have been only a few percent slower than the the single-set hash table implementation), but then I thought I don't need to since its not relevant for the competition result.

I think calling into the baseline class is OK? If not, I could just copy the code of it into my class, but that wouldn't change anything at the implementation, just code duplication.

prepare_kaufco.sh
kaufco: Fixed.

src/main/java/dev/morling/onebrc/CalculateAverage_kaufco.java
gunnarmorling: Each entry should be self-contained, i.e. calling out to the baseline should not be done. Would be nice to have a fallback which is reasonably fast. Keep in mind the goal is not to solely focus on the example data set but implement something which works in general for all allowed data sets.

calculate_average_kaufco.sh
gunnarmorling: Ah, ok. I suppose it's unmapping the file which takes a while. See latest (currently pending) entry of @thomaswue for dealing with that in a more graceful way.

src/main/java/dev/morling/onebrc/CalculateAverage_kaufco.java
kaufco: Fixed, see latest commit. I tested it with 100m entries @10k entries:
- execution time: 1.63s on my machine
- validated against output from merykitty

calculate_average_kaufco.sh
kaufco: Found it. Works. Thanks!

gunnarmorling:
@kaufco, could you point me to the place where you handle hash collisions? It's still a bit unclear to me. Thx!

gunnarmorling:
Thx for the explanation. As per the rules, there must be no potential for collisions _whatsoever_ (as _any_ 10K station names must be supported). So some means of resolving that collision must be implemented.

kaufco:
My understanding of the rules allowed for optimization steps that skip the full string comparison, provided there were no hash collisions for any of the provided input sets (i.e., that a full string comparison was not explicitly stated to be required). This was one of the fundamental aspects of my implementation which I cannot change. Unfortunately, if this is not in line with the challenge's requirements, I need to withdraw. Thank you for the opportunity.

gunnarmorling:
Sorry about that. This specific point is described in the [FAQ](https://github.com/gunnarmorling/1brc/#faq) in the README:

A: No, while only a fixed set of station names is used by the data set generator, any solution should work with arbitrary UTF-8 station names (for the sake of simplicity, names are guaranteed to contain no ; or \n characters).

Nevertheless thanks a lot for participating!

PR #529, opened 2024-01-21
CalculateAverage_michaljonko

michaljonko:
#### Check List:
- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number) - _Line 114_

* Execution time: ~2,9s
* Execution time of reference implementation: ~2m
* Machine: Intel i9-12900HK, 64GB Ram, PM9A1 NVMe


src/main/java/dev/morling/onebrc/CalculateAverage_michaljonko.java
```diff
@@ -0,0 +1,284 @@
...
+                stationNameHash = 31 * stationNameHash + b;
+            }
+
+            while (memorySegmentOffset < memorySegmentSize && (b = memorySegment.get(ValueLayout.JAVA_BYTE, memorySegmentOffset++)) != '\n') {
+                temperatureRaw[temperatureRawIndex++] = b;
+            }
+
+            final var temperature = TemperatureParser.parse(temperatureRaw, temperatureRawIndex);
+            stationMeasurement = stationsMap.get(stationNameHash);
+            if (stationMeasurement != null) {
```
japplis, line 115: I think that if 2 different station names would have the same hash, the first in the map would get the temperature of the other station.

gunnarmorling: Hum, yeah, indeed. You can't condense a 100 bytes string to an integer in a map without losing information.

gunnarmorling (changes requested):
Seems prone to hash collisions.

src/main/java/dev/morling/onebrc/CalculateAverage_michaljonko.java
michaljonko: For a case with known 10k cities (`weather_stations.csv`) it works. The last commits introduced a solution to minimize collisions when 10000 different city names are used.

michaljonko:
@gunnarmorling do you agree with the new solution?

gunnarmorling:
Unless I miss something, it still isn't guaranteed that all hash collisions are resolved, right? A valid entry must work correctly for _all_ valid station names.

gunnarmorling:
Gonna close this one for now as I haven't heard back and need to clean the slate a bit. Please feel free to re-open when you feel the time has come. Thx!

PR #530, opened 2024-01-21
Label unsafe entries

ianopolousfast:
Label entries using unsafe

I haven't got an update to evaluate.sh yet. But I'm happy to keep doing this manually at this stage of the comp, until/if I figure out a script update?

Also fix a source link (roman-r-m)


gunnarmorling:
Thanks a lot! I think an automated approach would be better though, as otherwise the info might get easily lost again for subsequent updates. @hundredwatt might perhaps have some ideas for automating that part? It's probably gonna be somewhat ugly:

* grep the class name from the launch script
* grep that class' source file for `sun.misc.Unsafe`; if it's present add note of that to the entry

japplis:
If it helps, beginning of this afternoon I wrote a program to see which implementation were what I would consider clean code (and fast). So here is the code.
I didn't grep on sun.misc.Unsafe as people could use sun.misc.*; also saw some comments like "didn't use unsafe". That's why I decided to grep on "theUnsafe".

public class BRCClean {

    private static final String BRC_PATH = "C:\\Users\\Anthony\\Dropbox\\Projects\\1brc";
    private static final String SOURCE_PATH = "src\\main\\java\\dev\\morling\\onebrc";
    private static final List<String> EXCLUDES = List.of(
            "arjenw", // Hash collision
            "gabrielreid", // 2 big methods
            "");

    public static void main(String... args) throws IOException {
        File sourceCode = new File(BRC_PATH + File.separator + SOURCE_PATH);
        List<String> cleanCode = new ArrayList<>();
        File[] javaFiles = sourceCode.listFiles((diir, name) -> name.startsWith("CalculateAverage_"));
        for (File javaFile : javaFiles) {
            String javaContent = Files.readString(javaFile.toPath());
            boolean isDirty = javaContent.contains("theUnsafe") ||
                    javaContent.contains(" << ") ||
                    javaContent.contains(" >> ") ||
                    javaContent.contains("& 0x");
            if (!isDirty) {
                String filename = javaFile.getName();
                String user = filename.substring(filename.indexOf("_") + 1, filename.lastIndexOf("."));
                cleanCode.add(user);
            }
        }
        cleanCode.removeAll(EXCLUDES);
        List<String> readmeLines = Files.readAllLines(Path.of(BRC_PATH, "README.md"));
        for (String user : cleanCode) {
            for (String readmeLine : readmeLines) {
                if (readmeLine.contains("/" + user) && (readmeLine.contains(" 00:0"))) { // || readmeLine.contains(" 00:1"))) {
                    System.out.println("> " + readmeLine);
                }
            }
        }
    }
}

gunnarmorling:
Yeah, we definitely could call out to Java, doesn't have to be a shell script if that makes things easier. No need to compile, can simply be called via `java find_the_unsafe.java`.

hundredwatt:
Wow, I have the 2nd fastest entry that's not using unsafe 😄 

I'll do a quick update to evaluate.sh for this... stay tuned

hundredwatt:
PR: #547

tivrfoa:
> If it helps, beginning of this afternoon I wrote a program to see which implementation were what I would consider clean code (and fast). So here is the code. I didn't grep on sun.misc.Unsafe as people could use sun.misc.*; also saw some comments like "didn't use unsafe". That's why I decided to grep on "theUnsafe".
> 
> public class BRCClean {
> 
> ```
> private static final String BRC_PATH = "C:\\Users\\Anthony\\Dropbox\\Projects\\1brc";
> private static final String SOURCE_PATH = "src\\main\\java\\dev\\morling\\onebrc";
> private static final List<String> EXCLUDES = List.of(
>         "arjenw", // Hash collision
>         "gabrielreid", // 2 big methods
>         "");
> 
> public static void main(String... args) throws IOException {
>     File sourceCode = new File(BRC_PATH + File.separator + SOURCE_PATH);
>     List<String> cleanCode = new ArrayList<>();
>     File[] javaFiles = sourceCode.listFiles((diir, name) -> name.startsWith("CalculateAverage_"));
>     for (File javaFile : javaFiles) {
>         String javaContent = Files.readString(javaFile.toPath());
>         boolean isDirty = javaContent.contains("theUnsafe") ||
>                 javaContent.contains(" << ") ||
>                 javaContent.contains(" >> ") ||

Nice! =)

If you're excluding shifts, you might as well exclude unsigned right shift >>>

gunnarmorling:
Hey @ianopolousfast, this got fixed now via https://github.com/gunnarmorling/1brc/pull/547 in an automated way, so we don't need the manual approach any more. Thanks a lot again for bringing this up an looking into it!

PR #536, opened 2024-01-21
Trying an aligned ByteBuffer (again) like Artsiom

royvanrijn:
This does give a much higher CPU utilisation on my old x86, perhaps unmap *is* the biggest bottleneck atm and this avoids it. Slightly more instructions using `perf`, but almost one core utilized extra, especially in the 8-core case this could make a difference.

#### Check List:

* Execution time: `00:01.86` (more on my M2)


gunnarmorling:
Hey, could you rebase and force-push this one? I've just expanded the GH Action for PRs and it should run tests against a PR's entry now. So let's see whether it works as expected :)

artsiomkorzun:
@royvanrijn now you should be trying @thomaswue approach, it will be much faster.

PR #538, opened 2024-01-21
Using the latest tricks

royvanrijn:
#### Check List:

First changed the code to use Artsiom's ByteBuffer, but I've noticed Thomas has an even better idea, so stealing that :-)

All other changes pretty much reverted.. I'd love to make the code simpler, but changing it to use a Predicate<byte[]> for updating the table makes things worse, which is kind of the opposite point of the challenge, so not doing that.

Perhaps a MethodHandle or something could be faster, don't know...

### DIFF
--- a/prepare_royvanrijn.sh
+++ b/prepare_royvanrijn.sh
@@ -16,7 +16,7 @@
 #
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
-sdk use java 21.0.1-graal 1>&2
+sdk use java 21.0.2-graal 1>&2
 
 # ./mvnw clean verify removes target/ and will re-trigger native image creation.
 if [ ! -f target/CalculateAverage_royvanrijn_image ]; then
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
@@ -15,12 +15,15 @@
  */
 package dev.morling.onebrc;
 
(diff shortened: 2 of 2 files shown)


prepare_royvanrijn.sh
```diff
@@ -16,7 +16,7 @@
 #
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
-sdk use java 21.0.1-graal 1>&2
+sdk use java 21.0.2-graal 1>&2
```
gunnarmorling, line 19: Wow, so cool to see how @hundredwatt's work is paying of: running _./evaluate.sh royvanrijn_ automatically installs that new version now, without anything to be done by hand. Noice!

gunnarmorling:
00:02.313 now, nice!

```
  Time (mean ± σ):      2.312 s ±  0.021 s    [User: 0.002 s, System: 0.003 s]
  Range (min … max):    2.270 s …  2.342 s    10 runs

Summary
  royvanrijn: trimmed mean 2.31353707266, raw times 2.27034278566,2.32108241566,2.29975882466,2.32893833366,2.30389421566,2.34196373966,2.30766307666,2.30471392566,2.33684425966,2.30540152966

Leaderboard

|   | 00:02.313 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java)| 21.0.2-graal | [Roy van Rijn](https://github.com/royvanrijn) | GraalVM native binary |
```

PR #544, opened 2024-01-21
Add 1brc solution by @makohn

makohn:
#### Check List:
- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions: 134

* Execution time: 5.184s (7.938s when limited to 8 cores)
* Execution time of reference implementation: 1m44.893s (1m49.146s when limited to 8 cores)

#### Environment:
- 14 Core Intel® Core™ i5-13500 (20 threads via HTT)
- 64 GB DDR4
- Ubuntu 22.04 LTS
- OpenJdk 21.0.1

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_makohn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_makohn.java
@@ -1,0 +1,287 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_makohn.java
```diff
@@ -0,0 +1,287 @@
...
+
+    // Twice as big as the maximum number of stations
+    private static final int MAP_CAPACITY = MAX_STATIONS * 2;
+
+    // We start at 1 to allow for checking our hash-index map for > 0
+    private static final int RES_FIRST_INDEX = 1;
+
+    private static class ResultMap {
+        final int[] map = new int[MAP_CAPACITY]; // hash -> index
+        final Measurement[] measurements = new Measurement[MAX_STATIONS]; // index -> measurement
```
gunnarmorling, line 108: I've seen this a couple of times, what's the advantage of having this indirection with two arrays, rather than probing within one?

makohn: That way, it's easier/faster to iterate over the `Measurement` array, since its entries are stored in order without gaps in between (and we know the number of actual entries as per `lastIndex`) (see lines 261 and 280)

If you would use the hash directly as an index, you need to filter out null/0 values later when merging the results or when building the output string.

Though, there might be some clever way to design a hash function to reduce/avoid the gaps, achieving a more dense distribution.

gunnarmorling: I see, yes, it can save you some iterations. Makes sense, thx!

PR #547, opened 2024-01-21
evaluate.sh: Add note for "using Unsafe"

hundredwatt:
Check for string `theUnsafe` in the java source file with: `grep -F "theUnsafe" -q ./src/main/java/dev/morling/onebrc/CalculateAverage_$fork.java`

Closes #530 

Tested with 4 cases:
* hundredwatt - base (no Unsafe or native image)
* merykittyunsafe - uses Unsafe; no native image
* Kidlik - uses native image; no unsafe
* royvanrijn - uses Unsafe and native image

Output:

```
1brc $ ./evaluate.sh hundredwatt merykittyunsafe Kidlike royvanrijn

...

Leaderboard

| 1 | 00:XX.XXX | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java)| 21.0.2-graal | [Roy van Rijn](https://github.com/royvanrijn) | GraalVM native binary, uses Unsafe |
| 2 | 00:XX.XXX | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_hundredwatt.java)| 21.0.1-graal | [Jason Nochlin](https://github.com/hundredwatt) |  |
| 3 | 00:XX.XXX | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java)| 21.0.1-open | [merykittyunsafe](https://github.com/merykittyunsafe) | uses Unsafe |
|   | 00:XX.XXX | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_Kidlike.java)| use | [Kidlike](https://github.com/Kidlike) | GraalVM native binary |

...
```


hundredwatt:
Backfill Ruby (was a bit more complicated than something I'd want to stick with bash for 😉):

```ruby
leaderboard = `tail -n+44 README.md | head -n 122`.split("\n")

leaderboard.each do |entry|
    begin
        file = entry.split("|")[3].match(/src.*[.]java/)[0]
        fork = File.basename(file, ".java").sub("CalculateAverage_", "")
        usesUnsafe = `grep -F theUnsafe #{file}`.length > 0
    
        notes = entry.split("|")[6]
        new_notes = notes.chop
        new_notes += ", " unless new_notes.match(/^ *$/)
        new_notes += "uses Unsafe" if usesUnsafe
        new_notes += " "
    
        new_entry = entry.split("|")
        new_entry[6] = new_notes
        puts new_entry.join("|") + "|"
    rescue
        puts entry
        fail
    end
end
```

hundredwatt:
Use `--word-diff` to easily see changes to Leaderboard:

```bash
git show --word-diff 1de1f9f -- README.md
```

evaluate.sh
```diff
@@ -271,6 +271,12 @@ for fork in "$@"; do
     fi
   fi
 
+  # check if Java source file uses Unsafe
+  if grep -F "theUnsafe" -q ./src/main/java*/dev/morling/onebrc/CalculateAverage_$fork.java ; then
```
gunnarmorling, line 275: This wouldn't work for those implementations where the fork name (i.e. GH account name) doesn't make for a valid Java class name (like in the case of `roman-s-m`). But that's alright, it doesn't matter that much in the large scheme of things.

gunnarmorling (approved):
Awesome, thank you so much, @hundredwatt!

Btw. as we speak: your own implementation currently doesn't pass the 10K key set evaluation. If you have the time/interest, it would be great to get it fixed, as it's part of the top 15 entries from the regular leaderboard which I occasionally run against that key set too.

hundredwatt:
@gunnarmorling Fixed my entry: #558  😄

PR #553, opened 2024-01-22
serkan-ozal: Initial impl

serkan-ozal:
#### Check List:
- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number): `583`

* Execution time: 3.7s
* Execution time of reference implementation: 5m 54.1s

**Here are configurations of my testing environment:**
| Name | Value |
| :---------------- | :------: | 
| Platform | AWS EC2 |
| OS Name | Amazon Linux |
| OS Version | 2023 |
| Kernel Version | Linux 6.1.66-91.160.amzn2023.x86_64 |
| Instance type | t3.2xlarge |
| CPU Architecture | x86_64 |
| CPU Model | Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz |
| Number of vCPUs | 8 |
Memory (GiB) | 32.0 |

# Notes:

- Like many others, my solution is also suffering from unmap delay (250-300 ms in my testing environment). But spawning worker Java process from main Java process (as suggested by `Thomas Wuerthinger`) in non-GraalVM solution doesn't help much, as it comes its new JVM process spin-up cost. So ,instead, (not sure it is allowed, but I think it is similar approach and can also be used by GraalVM based solutions too) in my startup script file (`calculate_average_serkan-ozal.sh`), I haved used `exec` to spawn the actual Java process to do all the stuff, redirect its stdout to a custom file descriptor, get the result from that file descriptor and complete the script execution without waiting real worker Java process as it deals with unmap.

### DIFF
--- a/calculate_average_serkan-ozal.sh
+++ b/calculate_average_serkan-ozal.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+#  Copyright 2023 The original authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+JAVA_OPTS="--enable-preview --enable-native-access=ALL-UNNAMED --add-modules=jdk.incubator.vector"
+JAVA_OPTS="$JAVA_OPTS -XX:-TieredCompilation -XX:MaxInlineSize=10000 -XX:InlineSmallCode=10000 -XX:FreqInlineSize=10000"
+JAVA_OPTS="$JAVA_OPTS -Djdk.incubator.vector.VECTOR_ACCESS_OOB_CHECK=0"
+JAVA_OPTS="$JAVA_OPTS -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -Xms256m -Xmx256m -XX:+AlwaysPreTouch"
+if [[ ! "$(uname -s)" = "Darwin" ]]; then
+  JAVA_OPTS="$JAVA_OPTS -XX:+UseTransparentHugePages"
+fi
+
+# TODO Generate/Get a random and available file descriptor number?
+fd=1234
+#echo "Process started at $(date +%s%N | cut -b1-13)"
+eval "exec $fd< <({ CLOSE_STDOUT_ON_RESULT=true USE_SHARED_ARENA=true java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_serkan_ozal; })"
+read <&$fd result
+echo -e "$result"
+#echo "Process finished at $(date +%s%N | cut -b1-13)"
--- a/calculate_average_serkan-ozal.sh
+++ b/calculate_average_serkan-ozal.sh
@@ -23,10 +23,8 @@ if [[ ! "$(uname -s)" = "Darwin" ]]; then
   JAVA_OPTS="$JAVA_OPTS -XX:+UseTransparentHugePages"
 fi
 
-# TODO Generate/Get a random and available file descriptor number?
-fd=1234
 #echo "Process started at $(date +%s%N | cut -b1-13)"
-eval "exec $fd< <({ CLOSE_STDOUT_ON_RESULT=true USE_SHARED_ARENA=true java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_serkan_ozal; })"
-read <&$fd result
+eval "exec 3< <({ CLOSE_STDOUT_ON_RESULT=true USE_SHARED_ARENA=true java $JAVA_OPTS --class-path target/average-1.0.0-SNAPSHOT.jar dev.morling.onebrc.CalculateAverage_serkan_ozal; })"
+read <&3 result
 echo -e "$result"
 #echo "Process finished at $(date +%s%N | cut -b1-13)"
--- a/calculate_average_serkan-ozal.sh
+++ b/calculate_average_serkan-ozal.sh
@@ -18,7 +18,7 @@
 JAVA_OPTS="--enable-preview --enable-native-access=ALL-UNNAMED --add-modules=jdk.incubator.vector"
 JAVA_OPTS="$JAVA_OPTS -XX:-TieredCompilation -XX:MaxInlineSize=10000 -XX:InlineSmallCode=10000 -XX:FreqInlineSize=10000"
 JAVA_OPTS="$JAVA_OPTS -Djdk.incubator.vector.VECTOR_ACCESS_OOB_CHECK=0"
-JAVA_OPTS="$JAVA_OPTS -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -Xms256m -Xmx256m -XX:+AlwaysPreTouch"
+#JAVA_OPTS="$JAVA_OPTS -XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -Xms256m -Xmx256m -XX:+AlwaysPreTouch"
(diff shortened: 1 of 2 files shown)


serkan-ozal:
Hi @gunnarmorling, 
Since this is my first PR, your approval is required (one time only) to run Github action workflow for PRs from me.

yavuztas:
Sounds cool, Serkan how much did you save? Any percentage? 

Well, this GraalVM is an interesting beast. Initially just switching to GraalVM was a %80 boost for me, without any code change. But at some point, it's getting weird. I was able to do some very clear-cut optimizations that save 50 - 100 ms but at some point GraalVM ignored them. Compiling to native is even worse. Maybe Graal doesn't like it when you optimize before it does.

I would like to try your method in GraalVM. But who knows how it performs, let's see :)

serkan-ozal:
@yavuztas 

It cut off all the unmap (which was about 250-300 ms in my case) delay at the end.

I am also thinking to test my solution with GraalVM, but first I think, I need to get rid of Vector API usage in my code. Because AFAIK, GraalVM doesn't properly handle intrinsifications for the Vector API yet.
Anyway, testing with GraalVM is on my radar, will see whether it helps.

serkan-ozal:
@gunnarmorling Gentle pinging. Had you a chance to review the code (or maybe test and run benchmark)?

yavuztas:
Hi @serkan-ozal, I realized lately after I had a look, that it's only the bash script which spawning the process. So, theoretically, it should work in any JVM. 

Anyway, the script didn't work for me on MacOS. Are you working on a Linux distro? I will try to adapt it for MacOS. Let's see.

serkan-ozal:
Hi @yavuztas,
My development env is MacOS but my final testing env is Linux. While testing on my local initially, I am not using that trick. I have not put effort for adapting it to MacOS and even not sure it is allowed in the 1brc challenge. Waiting for answer from Gunnar.

serkan-ozal:
@yavuztas 
BTW, apart from this topic, I have tested my solution in the GraalVM (by getting rid of Vector API). Initially it was almost 100% slower. Then, I made some modifications and now it is 20% slower than my original OpenJDK solution here. Will take another look whether I can make it better on GraalVM.

yavuztas:
@serkan-ozal 
Hmm, that explains why I cannot get any benefit when I try Thomaswue's solution on MacOS. Looks like it only makes a difference in Linux. Then, I need a Linux box to test :)

jerrinot:
it really depends. 
I'm testing on Linux and my solution is fastest on GraalVM with JIT, a bit slower with GraalVM AOT. and slowest with Hostpost C2. 
I haven't looked at the reasons why.

serkan-ozal:
@gunnarmorling 

Pinging again. Hope that I will get a response this time. As far as I see, my submission is the only submission which has not evaluated yet. Is it because of the approach I have mentioned in the notes or is there any specific reason.

I have other improvements on the way, but first I want to see how this works on the eval machine.

gunnarmorling:
@serkan-ozal, sorry for the delay, indeed the shell script approach was what made me put it a bit on the backburner. Just ran it on the eval instance and got this:

```
+ timeout -v 300 ./test.sh serkan-ozal
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-10000-unique-keys.txt
./calculate_average_serkan-ozal.sh: line 29: 1234: Bad file descriptor
./calculate_average_serkan-ozal.sh: line 30: $fd: Bad file descriptor
```

serkan-ozal:
@gunnarmorling it seems about the shell interpreter being used. I will check

serkan-ozal:
@gunnarmorling I have sent a commit which will fix the bad file descriptor issue hopefully. Could you please rerun it?

serkan-ozal:
@gunnarmorling 

I think it is related to file descriptor limit in the eval machine. Before the fix, I have used `1234` as the file descriptor number. So for example, if the file descriptor limit is `1024` in the eval machine, `1234` becomes invalid file descriptor number and fails. 

I have reproduced the same case by setting file descriptor limit to `1024` on my test env with `ulimit`. Anyway, with the latest commit, I have used `3` as the file descriptor and it should work.

gunnarmorling:
Ok, looking better, but running into an OOME now when testing the 1B file on 32 cores:

```
+ timeout -v 300 ./test.sh serkan-ozal
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-10000-unique-keys.txt
WARNING: Using incubator modules: jdk.incubator.vector
1c1,10000
< Terminating due to java.lang.OutOfMemoryError: Java heap space
---
> id1;1.0;1.0;1.0
```

serkan-ozal:
@gunnarmorling 

Sorry for the 2nd test fail. 

Even though I am not able to reproduce OOM, I believe that (if we get heap dump and open it) it is caused by short lived Vector API's `ByteVector` objects which are not collected by Epsilon GC. Maybe JIT behaves different in the eval machine and they are allocated on the heap.

Any way, not sure at the moment and as a fix attempt, I have just disabled Epsilon GC and rely on default GC. I will look into this latter but first just want to have a stabil working example.

Could you please rerun the tests?

gunnarmorling:
Hum, somehow this one is jinxed, LOL. There's a test failure now. It doesn't show up in the CI run, but it does so reproducibly on the eval machine:

```
+ timeout -v 300 ./test.sh serkan-ozal
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-10000-unique-keys.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-10.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-1.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-20.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-2.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-3.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-boundaries.txt
WARNING: Using incubator modules: jdk.incubator.vector
Validating calculate_average_serkan-ozal.sh -- src/test/resources/samples/measurements-complex-utf8.txt
WARNING: Using incubator modules: jdk.incubator.vector
15a16
> igButeboJuršinciKoaniImdinaNova VasDestrnikVarvarinSkomunGornji PetrovciRibnicaKon TumŠavnikPoul;22.5;22.5;22.5
17c18
< igButeboJuršinciKoaniImdinaNova VasDestrnikVarvarinSkopunGornji PetrovciRibnicaKon TumŠavnikPoul;18.5;20.5;22.5
---
> igButeboJuršinciKoaniImdinaNova VasDestrnikVarvarinSkopunGornji PetrovciRibnicaKon TumŠavnikPoul;18.5;18.5;18.5

FAILURE: ./test.sh serkan-ozal failed
Summary
  serkan-ozal: command failed or output did not match
```

serkan-ozal:
@gunnarmorling In my solution, hashes of those city names are same, so there are hash collisions for them. But they are identified as different city names by key equality check. Not sure how they are identified as equal only at the eval machine and works correct all the other environments. Anyway, looking into this.

serkan-ozal:
Hi @gunnarmorling, is the test environment big endian?

gunnarmorling:
Nope, x86_64 is little endian.

serkan-ozal:
@gunnarmorling 
Not sure it is the same issue, but I have fixed a potential byte order caused bug in the key equals check logic. Let's see whether it fixes this issue too. Could you please rerun one more time :)

gunnarmorling:
Same failure as before.

serkan-ozal:
@gunnarmorling Ah sorry, will try to reproduce this issue first.

gunnarmorling:
Given it's just that particular test that fails, I'd suspect, it's some character encoding thing, in particular looking at those diacritics. But it's just a guess.

serkan-ozal:
@gunnarmorling I was able to reproduce similar issue with different number of threads in my local. Not sure it is the same problem though

serkan-ozal:
@gunnarmorling 
Me again :) Could you please start another run? Let's hope this time there will not be any test failure.

gunnarmorling:
Yepp, looking good now. 00:02.933, nice!

Issue #540, opened 2024-01-21
Cache SDKMan in CI build


gunnarmorling:
Via #104, we have support now for running tests against a PR's owner's submission. SDKMan is installed from scratch though each time, as is the SDK referenced in the prepare script of the entry (if present). It would be nice to cache this.

gunnarmorling:
@ascheman, this might be something for you to explore?

PR #554, opened 2024-01-22
#540 Cache SDKman

ascheman:
@gunnarmorling - my solution for #540.

IMPORTANT: Only use SDKman provided Java - System JDK no longer installed!


ascheman:
FInd some test runs over [here](https://github.com/ascheman/1brc/actions), @gunnarmorling.
I have only cleaned up after my tests before submitting the PR.

PR #562, opened 2024-01-22
Rewrote to always read 16 bytes, this has less instructions on perf.

royvanrijn:
It doesn't make a lot of sense since quite some code can be written shorter, but this is what gives the best numbers. At least the 4 versions which different checks is mostly gone, just two left. Trying to make the code more predictable for the CPU.

Went from +/- 86000000000 cycles to 77000000000 on an old Intel MacBook; this has proven to be a better indicator than my usual M2.

#### Check List:

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java
@@ -62,7 +62,10 @@ import sun.misc.Unsafe;
  * Unrolling scan-loop:              1200 ms (seems to help, perhaps even more on target machine)
  * Adding more readable reader:      1300 ms (scores got worse on target machine anyway)
  *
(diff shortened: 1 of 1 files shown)


gunnarmorling:
A very nice improvement indeed!

```
  Time (mean ± σ):      2.194 s ±  0.022 s    [User: 0.001 s, System: 0.004 s]
  Range (min … max):    2.148 s …  2.218 s    10 runs

Summary
  royvanrijn: trimmed mean 2.19637615786, raw times 2.14809864036,2.19237090436,2.21775778136,2.21420512236,2.18576224736,2.21025219536,2.17868145436,2.19132771136,2.21686475936,2.18154486836

Leaderboard

|   | 00:02.196 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_royvanrijn.java)| 21.0.2-graal | [Roy van Rijn](https://github.com/royvanrijn) | GraalVM native binary, uses Unsafe |
```

PR #568, opened 2024-01-23
Use simd for name comparison

ianopolousfast:
Not sure if this will be faster on the eval machine, but it's much simpler and uses simd for name comparison. 

If it's slower then don't merge.

Hash collisions handled on line 120.

#### Check List:

* Execution time: 16s


gunnarmorling:
00:06.670 now.

PR #575, opened 2024-01-24
Adding solution for cb0s

cb0s:
#### Check List:

* Execution time:

#### Performance

On a AMD Ryzen 9 7950X with 64GB DDR5-CL28-5800MHz RAM and a PCIe Gen4 SSD, my algorithm completed in ~2.0s+/-0.1s on 32 threads. I compared it to thomas_wue's solution which runs in a little less than one third of the time that's why I hope I will be below 10 sec. on the testing machine without using anything like Unsafe or applying crazy bitmasks for less comparisons.

### DIFF
--- a/prepare_cb0s.sh
+++ b/prepare_cb0s.sh
@@ -16,13 +16,5 @@
 #
 
 # Uncomment below to use sdk
-source "$HOME/.sdkman/bin/sdkman-init.sh"
-sdk use java 21.0.1-graal 1>&2
-
-# used from thomaswue's prepare
-if [ ! -f target/CalculateAverage_cb0s_image ]; then
-    NATIVE_IMAGE_OPTS="--gc=epsilon -O3 -march=native --enable-preview"
-    # Use -H:MethodFilter=CalculateAverage_cb0s.* -H:Dump=:2 -H:PrintGraph=Network for IdealGraphVisualizer graph dumping.
-    native-image $NATIVE_IMAGE_OPTS -cp target/average-1.0.0-SNAPSHOT.jar -o target/CalculateAverage_cb0s_image dev.morling.onebrc.CalculateAverage_cb0s
-fi
-
+# source "$HOME/.sdkman/bin/sdkman-init.sh"
+# sdk use java 21.0.1-graal 1>&2
--- a/prepare_cb0s.sh
+++ b/prepare_cb0s.sh
@@ -20,7 +20,6 @@ source "$HOME/.sdkman/bin/sdkman-init.sh"
 sdk use java 21.0.1-graal 1>&2
 
 # used from thomaswue's prepare
-./mvnw clean verify # removes target/ and will re-trigger native image creation.
 if [ ! -f target/CalculateAverage_cb0s_image ]; then
     NATIVE_IMAGE_OPTS="--gc=epsilon -O3 -march=native --enable-preview"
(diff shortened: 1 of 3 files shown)


prepare_cb0s.sh
```diff
@@ -0,0 +1,29 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# Uncomment below to use sdk
+source "$HOME/.sdkman/bin/sdkman-init.sh"
+sdk use java 21.0.1-graal 1>&2
+
+# used from thomaswue's prepare
+./mvnw clean verify # removes target/ and will re-trigger native image creation.
```
gunnarmorling, line 23:
```suggestion
```

Not needed here.

gunnarmorling:
Hey @thomaswue, I am repeatedly running into this error when executing ths one for the 1B rows file:

```
Validating calculate_average_cb0s.sh -- measurements_1B.txt
Picking up existing native image 'target/CalculateAverage_cb0s_image', delete the file to select JVM mode.
Fatal error: Failed to leave the current IsolateThread context and to detach the current thread. (code 12)
```

Any idea what could be causing this? Couldn't find anything relevant so far.

gunnarmorling:
So as stated above, this fails in native mode, it's not clear to me what's happening. It passes on the JVM in 00:10.284. Happy to add that to the leaderboard for the time being, I am not sure how to resolve the issue with the native GraalVM binary. Let me know if you'd like to have the JVM result added.

wirthi:
Note: we are tracking that `Fatal Error` in GraalVM as ticket GR-51694 and try to reproduce it.

thomaswue:
@cb0s This could be related to the process running out of memory before shutdown. I believe you are running with epsilon GC. Maybe there is too much allocation and a collecting GC would be required?

cb0s:
Sorry for the late response. We currently have exams at uni...

I would love if you, @gunnarmorling , could add my JVM result to the leaderboard for the time being. I have to admit, it's the first time I am playing around with GraalVM. I just tested it on my system again, with epsilon GC (thanks @thomaswue for the suggestion) it runs fine. I will commit the changes to the running script now, I hope then it should also run on the test runner.

cb0s:
I did some more digging... Yes @thomaswue I am using the epsilon GC. When I used another GC it worked again. I believe, when I tested it, I must have been using a different GC mechanism. Funny enough, the pure Java solution is faster than the GraalVM solution. I added some more arguments making this solution a tiny bit faster. I will commit the changes now, and then this PR should be ready to merge @gunnarmorling . :)

cb0s:
Some additional stuff I noticed while testing.
1. CPU isn't fully utilized sometimes (I think the IO reading is a bottleneck and more direct read process (i.e. through unsafe or something) could improve the performance quite a bit.
2. I think the more interesting part for those trying to reproduce the GraalVM error: for quicker and easier comparison and hashing, every time I parsed one line, I create a RawName record - a wrapper for the underlying byte[]-array. These however (obviously) are not cleaned up afterwards, that's probably why epsilon GC won't work. An idea (which I am not sure, if I can test before the end of this contest) would be to make the wrapper class obsolete and therefore having a smaller memory footprint.

gunnarmorling:
All looking good now, 00:13.729 now with the JVM. Good luck for your exams :)

PR #579, opened 2024-01-24
CalculateAverage_gonix update

gonix:
Minor updates here and there, shaves off ~5% of execution time on my machine.

#### Check List:
  - See lines 208 and 239

* Execution time: ~1.75s
* Execution time of reference implementation: ~145s
* Machine: Intel i7-11800H @ 2.30GHz (max 4.6GHz), 8 cores (16 threads), 32GB RAM

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_gonix.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_gonix.java
@@ -79,13 +79,13 @@ public class CalculateAverage_gonix {
 
 class Aggregator {
     private static final int MAX_STATIONS = 10_000;
-    private static final int MAX_STATION_SIZE = (100 * 4) / 8 + 5;
+    private static final int MAX_STATION_SIZE = Math.ceilDiv(100, 8) + 5;
     private static final int INDEX_SIZE = 1024 * 1024;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_gonix.java
```diff
@@ -205,27 +205,15 @@ public Stream<Entry> stream() {
 
     private void add(ByteBuffer buf, int start, long tailAndLen, int hash, int measurement) {
         int idx = hash & INDEX_MASK;
-        while (true) {
-            if (index[idx] != 0) {
-                int offset = index[idx];
-                if (keyEqual(offset, buf, start, tailAndLen)) {
```
gunnarmorling, line 211: How do you deal with key collisions now?

gonix: The same as before: by comparing the key. I've moved entry update logic into the `keyEqual()` method and renamed it to `update()`, which returns `false` if the keys are not equal (collision), in which case the search continues to the next slot.

gunnarmorling: Ah yes, got it. Thx!

gunnarmorling:
00:03.824 now.

PR #580, opened 2024-01-24
Updates for gamlerhart: Simpler & Faster

gamlerhart:
#### Check List:

On my 6 core / 12thread laptop
* Execution time: 5-6 seconds
* Execution time before: 7-8 seconds


gamlerhart:
For the hash collisions, nothing new, but I never pointed it out:

Its handled in `doAdd`. There is the loop around the probe counter, which is increased on a non-matching collision.
It probes the map linearly.

Not implemented is any hash-map resizing

gamlerhart:
PS: Also updates the rounding bugfix.

gunnarmorling:
00:05.705 now.

PR #582, opened 2024-01-24
tonivade improved solution

tonivade:
#### Check List:
- [X] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (method starting at line number 202) 

This is an improved version of my previous solution. This time I removed the usage of HashMap and I'm using a two dimensions array to store the stations.

I calculate the hash code of the name  and then I apply the modulus to array length to get the bucket to store the station. If there are more than one station on the bucket, I iterate searching for the station with the same name length and hash code. 

I'm not sure if this makes the solution potential for hash collisions, but it passes all the tests and the solution matches with the baseline output.

* Execution time:
```
real    0m9,313s
user    1m8,684s
sys     0m1,613s
```
```
real    3m48,755s
user    3m45,781s
sys     0m3,594s
```

Executed in my laptop:
- Intel Core i7-1065G7
- 32GB RAM
- 4 cores and 2 threads per core

### DIFF
--- a/calculate_average_tonivade.sh
+++ b/calculate_average_tonivade.sh
@@ -15,5 +15,5 @@
 #  limitations under the License.
 #
 
-JAVA_OPTS="-Xmx1G -Xms1G -XX:+AlwaysPreTouch --enable-preview"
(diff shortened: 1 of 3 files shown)


gunnarmorling:
00:07.894 now.

PR #590, opened 2024-01-26
First attempt with Java-managed concurrency

dqhieuu:
#### Check List:

* Execution time: 82s on Macbook Pro M1
* Execution time of reference implementation: 300s
* Execution time of artsiomkorzun: 24s / For reference, 2.019s on your server

Kinda curious how much faster it will run on a server with much better parallelism than my current PC.

### DIFF
--- a/calculate_average_dqhieuu.sh
+++ b/calculate_average_dqhieuu.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+#
+#  Copyright 2023 The original authors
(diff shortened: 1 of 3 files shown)


gunnarmorling:
As per the README and the PR template, new entries should run in ~10 sec or less, this one is at 65. Still merging it as you put in the effort already, but any potential follow-up PRs should be somewhere up there.

PR #604, opened 2024-01-27
Initial submission for varunsankethi

varunsankethi:
#### Check List:
- [x ] Tests pass (`./test.sh <username>` shows no differences between expected and actual outputs)

* Execution time: 1 minutes 45 seconds
* Execution time of reference implementation: 2 minutes 25 seconds


gunnarmorling:
Hey, as per the README and the PR template, new entries at this point should run in about 10 sec on the eval machine, so as to keep things manageable somewhat. It's not a hard and fast rule, but as per the time you share and from a quick glance at the implementation this does seem way closer to the original base line. Do you think you could it improve it to get somewhere up there? If not, I'd rather we close this PR at this point. Thanks a lot!

varunsankethi:
Thanks for the feedback, I will try to reduce the running time and raise another PR in case it is under 10 seconds. Closing this PR.

gunnarmorling:
Thank you, appreciate it! Thanks a lot for participating in any case.

PR #605, opened 2024-01-27
Fix hash code collisions

jgrateron:
pass all test
time 16 sec
Ryzen 5 3GHZ 16GB RAM


gunnarmorling:
Thx! This looks better now indeed. It's a bit slower (00:22.457) but correct now, so I've merged it. Thx!

PR #510, opened 2024-01-20
Second attempt with various improvements

JamalMulla:
#### Check List:
- [x] Tests pass (`./test.sh <username>` shows no differences between expected and actual outputs)
- [x] All formatting changes by the build are committed
- [x] Your launch script is named `calculate_average_<username>.sh` (make sure to match casing of your GH user name) and is executable
- [x] Output matches that of `calculate_average_baseline.sh`
- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number)

* Execution time: 3s
* Execution time of reference implementation: 4m

Shaved off around 1.5s from my initial submission. Also trying with graal native binary as it seems to give slightly better results and all the top entries are also using it.

Most of the changes are just reducing function calls, inlining the hashing, etc

PR #479, opened 2024-01-18
GraalVM,Unsafe correct implementation,Optimized for L3 Cache Size

karthikeyan97:
#### Check List:
- [x] Tests pass (`./test.sh <username>` shows no differences between expected and actual outputs)
- [x] All formatting changes by the build are committed
- [x] Your launch script is named `calculate_average_<username>.sh` (make sure to match casing of your GH user name) and is executable
- [x] Output matches that of `calculate_average_baseline.sh`
- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number)

* Execution time:1200ms
* Execution time of reference implementation:208000ms

PR #606, opened 2024-01-27
Some fine tuning for thomaswue

thomaswue:
#### Check List:

* Execution time: 0.45s

This is some final (?) fine tuning. Should give a few more %. Next step would be to convert the hash table array into an "inlined data structure" with all unsafe access, but not sure if I want to go there ;-).

Update: Got this down to 0.41s  now on my machine with processing segments at 2MB each and also processing with 3 scanners in parallel in the same thread.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
@@ -29,10 +29,10 @@ import java.util.stream.IntStream;
  * Simple solution that memory maps the input file, then splits it into one segment per available core and uses
  * sun.misc.Unsafe to directly access the mapped memory. Uses a long at a time when checking for collision.
  * <p>
- * Runs in 0.60s on my Intel i9-13900K
+ * Runs in 0.45s on my Intel i9-13900K
  * Perf stats:
- *     34,716,719,245      cpu_core/cycles/
- *     40,776,530,892      cpu_atom/cycles/
+ *     33,128,048,027      cpu_core/cycles/
+ *     38,668,037,020      cpu_atom/cycles/
  */
 public class CalculateAverage_thomaswue {
     private static final String FILE = "./measurements.txt";
@@ -42,10 +42,11 @@ public class CalculateAverage_thomaswue {
     // Holding the current result for a single city.
     private static class Result {
         long lastNameLong, secondLastNameLong;
-        long[] name;
-        int count;
-        short min, max;
+        long min, max;
         long sum;
+        int count;
+        long[] name;
+        String nameAsString;
 
         private Result() {
             this.min = MAX_TEMP;
@@ -73,13 +74,16 @@ public class CalculateAverage_thomaswue {
         }
 
         public String calcName() {
-            ByteBuffer bb = ByteBuffer.allocate(name.length * Long.BYTES).order(ByteOrder.nativeOrder());
-            bb.asLongBuffer().put(name);
-            byte[] array = bb.array();
-            int i = 0;
-            while (array[i++] != ';')
-                ;
-            return new String(array, 0, i - 1, StandardCharsets.UTF_8);
+            if (nameAsString == null) {
+                ByteBuffer bb = ByteBuffer.allocate(name.length * Long.BYTES).order(ByteOrder.nativeOrder());
+                bb.asLongBuffer().put(name);
+                byte[] array = bb.array();
+                int i = 0;
+                while (array[i++] != ';')
+                    ;
+                nameAsString = new String(array, 0, i - 1, StandardCharsets.UTF_8);
+            }
+            return nameAsString;
         }
     }
 
@@ -95,9 +99,10 @@ public class CalculateAverage_thomaswue {
         // Parallel processing of segments.
         List<List<Result>> allResults = IntStream.range(0, chunks.length - 1).mapToObj(chunkIndex -> parseLoop(chunks[chunkIndex], chunks[chunkIndex + 1]))
                 .map(resultArray -> {
-                    List<Result> results = new ArrayList<>();
+                    List<Result> results = new ArrayList<>(500);
                     for (Result r : resultArray) {
                         if (r != null) {
+                            r.calcName();
                             results.add(r);
                         }
                     }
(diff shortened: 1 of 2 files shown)


gunnarmorling:
It doesn't seem to make a difference on the eval machine:

```
Using java version 21.0.2-graal in this shell.
Validating calculate_average_thomaswue.sh -- measurements_1B.txt
Picking up existing native image 'target/CalculateAverage_thomaswue_image', delete the file to select JVM mode.

+ rm -f measurements.txt
+ ln -s measurements_1B.txt measurements.txt
Benchmark 1: timeout -v 300 ./calculate_average_thomaswue.sh 2>&1
  Time (mean ± σ):      2.189 s ±  0.043 s    [User: 0.002 s, System: 0.004 s]
  Range (min … max):    2.075 s …  2.225 s    10 runs

Summary
  thomaswue: trimmed mean 2.19823098707, raw times 2.07534146632,2.19937587432,2.16664367332,2.2091708203200002,2.22455168132,2.22013504532,2.19476398532,2.20274883332,2.20429925032,2.18871041432

Leaderboard

|   | 00:02.198 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java)| 21.0.2-graal | [Thomas Wuerthinger](https://github.com/thomaswue) | GraalVM native binary, uses Unsafe |
```

gunnarmorling:
Btw. this might be interesting to you: I am seeing this exception on two other entries:

```
Fatal error: Failed to leave the current IsolateThread context and to detach the current thread. (code 12)
```

For instance [here](https://github.com/gunnarmorling/1brc/pull/510). Any idea what might be the cause and how to avoid it?

tivrfoa:
> Btw. this might be interesting to you: I am seeing this exception on two other entries:
> 
> ```
> Fatal error: Failed to leave the current IsolateThread context and to detach the current thread. (code 12)
> ```
> 
> For instance [here](https://github.com/gunnarmorling/1brc/pull/510). Any idea what might be the cause and how to avoid it?

It seems it comes from here:

https://github.com/oracle/graal/blob/9d05067f874b7c00f75b7f62cfea4a64eca93e33/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/c/function/CEntryPointSetup.java#L82

```java
    public static final class LeaveEpilogue implements CEntryPointOptions.Epilogue {
        private static final CGlobalData<CCharPointer> errorMessage = CGlobalDataFactory.createCString(
                        "Failed to leave the current IsolateThread context.");

        @Uninterruptible(reason = "epilogue")
        static void leave() {
            int code = CEntryPointActions.leave();
            if (code != CEntryPointErrors.NO_ERROR) {
                CEntryPointActions.failFatally(code, errorMessage.get());
            }
        }
    }
```

I'll try to find the cause.

Related problem:
https://github.com/quarkusio/quarkus/issues/6606#issuecomment-586340877

tivrfoa:
> . (code 12)

Maybe @peter-hofer can help =)

thomaswue:
Yes, this looks like a potential native image bug. I will check with the team.

tivrfoa:
> and to detach the current thread.

It's actually here (GitHub text search is not that good ...):

https://github.com/oracle/graal/blob/9d05067f874b7c00f75b7f62cfea4a64eca93e33/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/c/function/CEntryPointSetup.java#L95

```java
    public static final class LeaveDetachThreadEpilogue implements CEntryPointOptions.Epilogue {
        private static final CGlobalData<CCharPointer> errorMessage = CGlobalDataFactory.createCString(
                        "Failed to leave the current IsolateThread context and to detach the current thread.");

        @Uninterruptible(reason = "epilogue")
        public static void leave() {
            int code = CEntryPointActions.leaveDetachThread();
            if (code != 0) {
                CEntryPointActions.failFatally(code, errorMessage.get());
            }
        }
    }
```

tivrfoa:
> Yes, this looks like a potential native image bug. I will check with the team.

There's an open issue from Mar 24, 2021 for that same error msg and code:
https://github.com/oracle/graal/issues/3305

tivrfoa:
> Btw. this might be interesting to you: I am seeing this exception on two other entries:
> 
> ```
> Fatal error: Failed to leave the current IsolateThread context and to detach the current thread. (code 12)
> ```
> 
> For instance [here](https://github.com/gunnarmorling/1brc/pull/510). Any idea what might be the cause and how to avoid it?

There are at least three entries where this error happened:

https://github.com/gunnarmorling/1brc/pull/479

https://github.com/gunnarmorling/1brc/pull/510

https://github.com/gunnarmorling/1brc/pull/575

thomaswue:
@gunnarmorling Could you please evaluate this new version? It consistently beats the top entry now on my machine, so with a bit of luck, it could get below 2s ;-).

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
```diff
@@ -23,16 +23,17 @@
...
  * Simple solution that memory maps the input file, then splits it into one segment per available core and uses
  * sun.misc.Unsafe to directly access the mapped memory. Uses a long at a time when checking for collision.
  * <p>
- * Runs in 0.60s on my Intel i9-13900K
+ * Runs in 0.41s on my Intel i9-13900K
  * Perf stats:
- *     34,716,719,245      cpu_core/cycles/
- *     40,776,530,892      cpu_atom/cycles/
+ *     25,286,227,376      cpu_core/cycles/
+ *     26,833,723,225      cpu_atom/cycles/
```
gunnarmorling, line 36: Wow, huge diff. Didn't expect this at this point.

gunnarmorling:
It is below 2 sec indeed, and quite a bit in fact! Could you describe perhaps on a high level what is making this huge difference at this point?

```
Benchmark 1: timeout -v 300 ./calculate_average_thomaswue.sh 2>&1
  Time (mean ± σ):      1.890 s ±  0.011 s    [User: 0.002 s, System: 0.004 s]
  Range (min … max):    1.859 s …  1.898 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  thomaswue: trimmed mean 1.89342727442, raw times 1.85925432142,1.8941132214199998,1.8928165554199998,1.89327190942,1.89163260842,1.89773446642,1.89233219542,1.89314723642,1.89674720142,1.89335726742

Leaderboard

|   | 00:01.893 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java)| 21.0.2-graal | [Thomas Wuerthinger](https://github.com/thomaswue) | GraalVM native binary, uses Unsafe |
```

thomaswue:
Excellent! Below 2s. Mission accomplished ;-).

So the main changes are to adopt processing segments in 2MB blocks to make sure all threads finish at the same time (otherwise other threads might sit idle while one thread finishes its part). This is inspired by @artsiomkorzun's solution. Looking at the perf flamegraph of his solution showed the benefit in wall clock time.

The other major change is to process 3 scanners in parallel in the same thread. This increases instructions per cycle, because there is more straight-line code that the Intel processor can then optimize better. The machine code generation is also more efficient, because there are quite some 64-bit long constants involved (in finding delimiters) and those are better shared if the same operation is done 3 at a time.

gunnarmorling:
Awesome, really cool!

tivrfoa:
> Excellent! Below 2s. Mission accomplished ;-).
> 
> So one main change is to adopt processing segments in 2MB blocks to make sure all threads finish at the same time (otherwise other threads might sit idle while one thread finishes its part). This is inspired by @artsiomkorzun's solution. Looking at the perf flamegraph of his solution showed the benefit in wall clock time.
> 
> The other major change is to process 3 scanners in parallel in the same thread. This increases instructions per cycle, because there is more straight-line code that the Intel processor can then optimize better. The machine code generation is also more efficient, because there are quite some 64-bit long constants involved (in finding delimiters) and those are better shared if the same operation is done 3 at a time.

Beautiful solution Thomas! :star_struck: :star_struck: 

I think the main speed up came from smaller segments, and **doing that inside the hot loop !**, because you get the huge benefit of still using the same `Result[] results = new Result[1 << 17];`

I'm not sure how much performance was gained from unrolling (spliting) the scanners.
But you for sure knows!!! =)

thomaswue:
It is a bit difficult to isolate (and might be different on the benchmark machine), but from the total 15% improvement over the previous entry, it seems almost an equal contribution from both of those changes. Doing 3 in parallel definitely helps more than 5% on my machine.

src/main/java/dev/morling/onebrc/CalculateAverage_thomaswue.java
```diff
@@ -139,149 +159,228 @@ private static TreeMap<String, Result> accumulateResults(List<List<Result>> allR
...
-                    if (namePos >= existingResult.name.length || existingResult.name[namePos++] != scanner.getLongAt(nameAddress + i)) {
-                        tableIndex = (tableIndex + 31) & (results.length - 1);
-                        continue outer;
-                    }
+        // Final calculation for index into hash table.
+        int tableIndex = hashToIndex(hash, results);
+        outer: while (true) {
+            existingResult = results[tableIndex];
+            if (existingResult == null) {
+                existingResult = newEntry(results, nameAddress, tableIndex, nameLength, scanner);
```
tivrfoa, line 237: Wouldn't it be better if you `return newEntry...` here?

PR #608, opened 2024-01-28
jhonesto: first submission

jhonesto:
#### Check List:
- [] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (line number) **There is no custom hash structure here**

* Execution time: 4.606 s
* Execution time of reference implementation: 3m20.823s

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_jhonesto.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_jhonesto.java
@@ -1,0 +1,395 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package dev.morling.onebrc;
+
+import java.io.*;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+
+public class CalculateAverage_jhonesto {
+
+    // +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+    // ++++++++++++++++++++++++++++ CONSTANTS +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+    // +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+    private static final String FILE = "./measurements.txt";
+    private static final String READ = "r";
+
+    private static final byte NEW_LINE = 10;
+    private static final byte MOD_NUMBER = 48;
+    private static final byte SEMICOLON = 59;
+    private static final byte MAX_PER_LINE = 107; // 107 = 100 station + ;-99.9 + \n
+
+    private static final int ZERO = 0;
+    private static final int ONE = 1;
+    private static int MAX_BUFFER_SIZE = 2_147_483_647 - 200;
+
+    private static final short RADIX = 10;
+    private static final short MIN_VALUE = -1000;
+    private static final short MAX_VALUE = 1000;
+
+    private static final long ZERO_LONG = 0L;
+
+    private static final double TEN_DOUBLE = 10.0d;
+
+    private static final char LINE_FEED = '\n';
+    private static final char MINUS = '-';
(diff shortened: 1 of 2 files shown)


src/main/java/dev/morling/onebrc/CalculateAverage_jhonesto.java
```diff
@@ -0,0 +1,395 @@
...
+
+        Measurement m = table.get(hash);
+
+        if (m != null) {
+            m.setTemperature(temperature);
+
+        }
+        else {
+            m = new Measurement(station, temperature);
+            table.put(hash, m);
```
gunnarmorling, line 374: Just using an integer hash as a key is prone to collisions. You can't reduce a 100 byte string to a 4 byte number without losing information.

jhonesto: There is no test case for this, if you reproduce a collision, I will think about it.

gunnarmorling: I can't create test cases for all potential hashing algorithms people may use. But it's clear that the entry is prone to hash collisions as-is and thus isn't valid. Let me know whether you'd like to rework it or retract it. Thx.

jhonesto: 1st -> I can't fix it if it's not broken.
2nd -> I'm using the Arrays.hashCode Algorithm used in the majority of implementations of this chalenge.
3rd -> A test case for collisions make more sense than a 1 row test.

There's two kind of situation here:

a - There's a collision "as-is". Source.: Voices of my mind.
b - There's a collision with stationA and stationB. 

Since my implementation passed all the tests, I will only fix situation B.

Thanks.

gunnarmorling: >  There's a collision "as-is". Source.: Voices of my mind

I don't appreciate your tone here. It's a standard procedure to point out issues during a code review, no test is needed when they are that apparent. It's surprising that we even debate this; if you could unambigiously express the information from 100 random bytes using four bytes, you could compress everything indefinitely.

> I'm using the Arrays.hashCode Algorithm used in the majority of implementations of this chalenge.

They either use keys which implement identity based on the actual station name (not just their hash), or they implement custom maps with open addressing, checking station name for equality in case of hash collisions.

> A test case for collisions make more sense than a 1 row test.

Again, how would that test look like? People use different functions and seed values so it's not doable on a general level. I might add sum fuzzy testing, but it would come at the cost of taking things even longer to run. Anyways, here's a patch you can apply which adds a test that fails with your implementation:

```
---
 src/test/resources/samples/measurements-hashcollision.out | 1 +
 src/test/resources/samples/measurements-hashcollision.txt | 2 ++
 2 files changed, 3 insertions(+)
 create mode 100644 src/test/resources/samples/measurements-hashcollision.out
 create mode 100644 src/test/resources/samples/measurements-hashcollision.txt

diff --git a/src/test/resources/samples/measurements-hashcollision.out b/src/test/resources/samples/measurements-hashcollision.out
new file mode 100644
index 00000000..42af246f
--- /dev/null
+++ b/src/test/resources/samples/measurements-hashcollision.out
@@ -0,0 +1 @@
+{pjhvyxbwwuifdgwhwoxz=19.2/19.2/19.2, wsyrjwjcdpmqeypcjlwh=9.5/9.5/9.5}
diff --git a/src/test/resources/samples/measurements-hashcollision.txt b/src/test/resources/samples/measurements-hashcollision.txt
new file mode 100644
index 00000000..dd34715c
--- /dev/null
+++ b/src/test/resources/samples/measurements-hashcollision.txt
@@ -0,0 +1,2 @@
+pjhvyxbwwuifdgwhwoxz;19.2
+wsyrjwjcdpmqeypcjlwh;9.5
-- 
```

```
./test.sh jhonesto src/test/resources/samples/measurements-hashcollision.txt
Validating calculate_average_jhonesto.sh -- src/test/resources/samples/measurements-hashcollision.txt
1c1,2
< pjhvyxbwwuifdgwhwoxz;9.5;14.4;19.2
---
> pjhvyxbwwuifdgwhwoxz;19.2;19.2;19.2
> wsyrjwjcdpmqeypcjlwh;9.5;9.5;9.5
```

jhonesto: I don't care if you appreciate my tone here. Think is cheap you need to prove your argument. 

Now I can treat this test accordingly. But, in my tests it's ok. There is no collision.

hash-collision

gunnarmorling:
Gonna close this one, as the implementation isn't valid and there's no intention to make the required changes.

jhonesto:
It's clear to me that you are unable to handle this situation and any argument presented here without facts is just hallucination.

I asked you to show me a collision test case, and you answered: "I cAn'T cReAtE a TeSt CaSe FoR eVeRy HaSh AlGoRiThM."

And I told you that the algorithm used Arrays.hashCode (that uses ArraysSupport.hashcode), is the same used in most implementations. We're not reinventing the wheel here, nor participating in a Cypherpunks contest. Therefore, it would only be a test case, with two lines demonstrating a collision. In this way, all implementations could be tested accordingly.

Now let's suppose that there is a collision, since you haven't shown me anything, if I increase the data type to long and use another more robust algorithm for hashing, and obviously passing the tests, will it be accepted by you?

Because at the end of the day I want to know if I'm participating in an One billion rows chalenge or wasting my time with the "SOME EGO INFLATED THAT DOESN'T KNOW HOW TO MAKE USEFULL TEST CASES." chalenge.

gunnarmorling:
Look, I have explained to you why this implementation isn't valid, I have given you a test case failing on the eval machine which proves this. Not sure what else you expect me to do. "Works on my machine" is not a convincing argument.

jhonesto:
You're right.

"It works on my machine" isn't an argument, and "it fails on my machine" isn't either. I'm trying to solve this problem.

Just answer my question with yes or no. Because I have read enough B.S. here.

Bye.

PR #609, opened 2024-01-28
Second version by albertoventurini

albertoventurini:
#### Check List:

* Execution time: < 10s
* Execution time of reference implementation: 3m19.421s

This version contains some minor optimisations in the parsing code.
Furthermore, I tested my previous version on an 8-core AMD EPYC machine and saw that the solution was performing particularly poorly on that environment (~30 seconds). I resized the byte buffers that each thread uses to read from the RandomAccessFile, and with this change, my solution runs under 10 seconds on the 8-core AMD machine.

### DIFF
--- a/calculate_average_albertoventurini.sh
+++ b/calculate_average_albertoventurini.sh
@@ -15,5 +15,5 @@
 #  limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:22.334 now.

PR #613, opened 2024-01-28
last perf papercut 

iziamos:
#### Check List:

* Execution time: just shy of 1 second on all cores.
* Execution time of reference implementation:  1:35.20

This was quite fun overall, learned alot. I'm just pushing this last tweak because I suspect its the lowest of the hanging fruit.

Farewell!

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_iziamos.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_iziamos.java
@@ -185,7 +185,7 @@ public class CalculateAverage_iziamos {
 
             byte b = UNSAFE.getByte(pointer);
             for (; b != ';'; ++strLen, b = UNSAFE.getByte(pointer + strLen)) {
-                hash += b << strLen;
+                hash = 31 * hash + b;
             }
             pointer += strLen + 1;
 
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Doesn't make a significant difference on the eval env: 00:04.167. More importantly though, the entry doesn't finish for the 10K key set test (see create_measurements3.sh and results [here](https://github.com/gunnarmorling/1brc#10k-key-set)). Could you take a look at this one?

iziamos:
Hmm worth a look, I suspect its the hashcode I'd switched to, for context I wanted something where only the position in the string mattered to see if it would be trivially vectorisable. It turned out that vectors were to slow though :(

iziamos:
Just did a squash, probably not worth pushing if it performs worse on the eval box but just to demonstrate that its a hash collision issue on the 10k. :)

gunnarmorling:
@iziamos, could you clarify: should this pass for 10K keys now, or is there still a potential hash collision? Correctness is the most important thing, so even if it is slower than what's on main right now but actually is more correct, I'd still merge it. Thx for clarifying.

iziamos:
@gunnarmorling I believe it was always correct, this version is significantly faster in the 10k keys (4 sec vs sth like 4 min), but slower (on my machine) for the main challenge.

In general I've found that the best way to evaluate solutions  for hash collisions isn't to stress test the key count but to replace the hashcode implementation with one that returns a constant e.g. 1, and see that it eventually returns correct results.

gunnarmorling:
Gotcha. So it's minimally slower for the standard key set (00:04.230), but indeed much faster for the 10K key set (00:06.712), so I'm gonna merge it, as this entry is amongst the top 20 or so for which I occassionally publish the 10K leaderboard and it DNF-ed there so far. Thx!

PR #623, opened 2024-01-28
Validation script to handle unsorted results and all results

ericxiao251:
I had written a validation script before knowing that such a script already existed upstream (https://github.com/gunnarmorling/1brc/blob/main/test.sh), but I noticed that the current testing would require both the ordering of the keys in the end maps of the fork to be the same as the main.

If desirable and my understanding is correct I can incorporate my script into the existing `test.sh` as my script might help determine the correctness of solutions that do not use an ordered map to store results. Additionally, this script also will check the full output and not just a sample.

### DIFF
--- a/ValidateResult.java
+++ b/ValidateResult.java
@@ -1,0 +1,68 @@
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hey @ericxiao251, thanks a lot for looking into this and sorry for the late reply. Results actually must be ordered by key as per the requirements ("sorted alphabetically by station name"), hence the existing _test.sh_ script doesn't do any sorting, as it expects the asserted output to be sorted correctly. In that light, I think we can close this PR. Thx again!

PR #627, opened 2024-01-28
shipilev: Amendments to version 4

shipilev:
#### Check List:
- [x] ~For new entries, or after substantial changes~ Not a new entry.

A few simple touchups after running with 8 cores:
 - Default FJP pool is one thread short when execution is capped at 8 cores;
 - No need for some JVM options in the constrained run;
 - Tiny locality improvements;

On standard dataset:

```
# Old impl:
Benchmark 1: ./calculate_average_shipilev.sh
  Time (mean ± σ):      3.819 s ±  0.091 s    [User: 25.632 s, System: 0.982 s]
  Range (min … max):    3.677 s …  3.990 s    10 runs

# New impl:
Benchmark 1: ./calculate_average_shipilev.sh
  Time (mean ± σ):      3.252 s ±  0.086 s    [User: 24.463 s, System: 0.906 s]
  Range (min … max):    3.196 s …  3.489 s    10 runs
```

So it should be about 15% faster.

### DIFF
--- a/calculate_average_shipilev.sh
+++ b/calculate_average_shipilev.sh
@@ -15,12 +15,11 @@
 #  limitations under the License.
 #
 
-JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseEpsilonGC -Xms64m -Xmx64m -XX:+AlwaysPreTouch -XX:+UseTransparentHugePages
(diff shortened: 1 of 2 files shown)


gunnarmorling:
00:04.884 now. As for your question on the original PR, the official eval is on 8 cores indeed, but I am doing an occasional side eval on all the cores, with SMT, for funsies.

PR #649, opened 2024-01-29
fix ordering in README.md

ianopolousfast:
Fix ordering


gunnarmorling:
Whoops, good catch, thx!

PR #653, opened 2024-01-29
anestoruk - one more improvement

anestoruk:
#### Check List:

- [x] For new entries, or after substantial changes: When implementing custom hash structures, please point to where you deal with hash collisions (`136-142`)

* Execution time (6-core i5 @ 3.70GHz): `Range (min … max):    6.970 s …  8.332 s    10 runs` 
* Execution time (4-core i5 @ 2.40GHz): `9-11s`

* Execution time of reference implementation: `~3m 4s`

Hello, one more improvement from me: city array byte buffer is copied only when new Record instance is created.

(there is also a second improvement - splitting the memory into a more even chunks - but this change will have no effect on a test that is run against 8-core machine)


gunnarmorling:
00:08.896 now.

PR #658, opened 2024-01-30
1brc submission by godofwharf

godofwharf:
#### Check List:

* Execution environment: 8 vcores AMD EPYC 7000 (Rome) / 16 GiB RAM / SMT disabled
* Execution time: 7.9 s
* Execution time of reference implementation: 3m 11s

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_godofwharf.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_godofwharf.java
@@ -1,0 +1,632 @@
+package dev.morling.onebrc;
+
+/*
+ *  Copyright 2023 The original authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


godofwharf:
@gunnarmorling Please review the same. My implementation doesn't use Unsafe and doesn't use Graal VM. It goes about processing the file in the following manner
- Maps the input file into memory
- Divides the input file into multiple splits and multiple pages
- Processes the multiple splits in parallel using a custom thread pool one page after another
- Uses Java's Vector API for identifying positions of new-line characters within each page
- Uses thread local state maps to perform local aggregation within each split
- Uses a ConcurrentHashMap to create a global aggregate map
- Sorts the station names by importing map entries into a tree map

src/main/java/dev/morling/onebrc/CalculateAverage_godofwharf.java
```diff
@@ -0,0 +1,632 @@
...
+        }
+
+        public void put(final State.AggregationKey key,
+                        final MeasurementAggregator aggregator) {
+            int h1 = key.hashCode;
+            int hash = h1 ^ (h1 >>> 12);
+            tableEntries[(size - 1) & hash] = new TableEntry(key, aggregator);
+        }
+
+        public MeasurementAggregator get(final State.AggregationKey key) {
```
gunnarmorling, line 603: How do you handle hash collisions here? Looks like you may end up aggregating the values for several stations with the same hash?

godofwharf: Hey @gunnarmorling, thanks for the review comment. I forgot to remove the FastHashMap implementation. It is not getting used because it results in collisions. I tried implementing double hashing for probing but the implementation proved to be slower than using Java HashMap. So I thought I would fix it in a future iteration. For now, I will remove this class because it is not getting used anywhere.

gunnarmorling: Ah got it. Thx for clarifying.

gunnarmorling:
00:11.363.

PR #659, opened 2024-01-30
batched writing to disk for randomized temperatures

nickyreinert:
instead of writing result line by line, implemented random.choices for randomisation of multiple stations and writing large batche ot the disk, also instead of "round" just using :.1f which is probably quicker on a large scale, because it's not a mathematical function

#### Check List:

**not applicable, this just improves generation of test data with Python**

- [ ] You have run `./mvnw verify` and the project builds successfully
- [ ] Tests pass (`./test.sh <username>` shows no differences between expected and actual outputs)
- [ ] All formatting changes by the build are committed
- [ ] Output matches that of `calculate_average_baseline.sh`

* Execution time:


gunnarmorling:
Thx!

PR #669, opened 2024-01-30
Final submission

merykitty:
Some small improvements:

- It is more performant to use branches than to use `min/max`
- Do 2 independent iterations to improve IPC

Some that I have tried:

- Do 4 iterations and use vector instructions to parse all of them at once: LongVector multiplication codegen is horrible, assembling a vector from its elements is expensive, also the constants need to be materialized for each iteration unless the compiler can inline the whole thing.
- Spawning a separate process to hide the cost of unmapping: It is likely that the cost of initializing the VM is higher than the cost of unmap, I did not see improvements.
- Some trickeries to avoid `Unsafe`: Spawning a new segment each time introduces regression, most likely due to failures to inline even with that threshold of inlining, and using a `Long.MAX_VALUE`-size segment introduces additional bound checks.

#### Check List:

* Execution time: Cannot tell, my laptop is not really reliable for benchmarking, but the elapsed time is around 20s.
* Execution time of the previous implementation: Around 26s.

### DIFF
--- a/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
+++ b/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java
@@ -227,11 +227,11 @@ public class CalculateAverage_merykitty {
 
         // Find the delimiter ';'
         long semicolons = line.compare(VectorOperators.EQ, ';').toLong();
-        int keySize = Long.numberOfTrailingZeros(semicolons);
 
         // If we cannot find the delimiter in the vector, that means the key is
         // longer than the vector, fall back to scalar processing
-        if (keySize == BYTE_SPECIES.vectorByteSize()) {
+        if (semicolons == 0) {
+            int keySize = BYTE_SPECIES.length();
             while (data.get(ValueLayout.JAVA_BYTE, offset + keySize) != ';') {
                 keySize++;
             }
@@ -240,6 +240,7 @@ public class CalculateAverage_merykitty {
         }
 
         // We inline the searching of the value in the hash map
+        int keySize = Long.numberOfTrailingZeros(semicolons);
         int x;
         int y;
         if (keySize >= Integer.BYTES) {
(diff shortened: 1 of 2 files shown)


tarsa:
> * Spawning a separate process to hide the cost of unmapping: It is likely that the cost of initializing the VM is higher than the cost of unmap, I did not see improvements.

it seems there's another trick to try:
https://github.com/gunnarmorling/1brc/blob/7f0e51781190bcfcd4b8624352230a19737b01c8/src/main/java/dev/morling/onebrc/CalculateAverage_serkan_ozal.java#L171-L176
along with
https://github.com/gunnarmorling/1brc/blob/7f0e51781190bcfcd4b8624352230a19737b01c8/calculate_average_serkan-ozal.sh#L31-L35

gunnarmorling:
Nice!

|   | 00:02.367 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_merykittyunsafe.java)| 21.0.1-open | [Quan Anh Mai](https://github.com/merykittyunsafe) | uses Unsafe |
|   | 00:03.210 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_merykitty.java)| 21.0.1-open | [Quan Anh Mai](https://github.com/merykitty) |  |

gunnarmorling:
Hey @merykitty, couldn't find another way to reach out to you, so commenting here: would you perhaps be interested in teaming up with Thomas, Roy, Elliot, and me for doing a live stream on #1BRC tomorrow with Nicolai Parlog: https://twitter.com/gunnarmorling/status/1752288991909478611. As one of the fastest JVM-based implementations, it would be great to have your perspective on the challenge. Let me know :)

merykitty:
@gunnarmorling I would be more than happy to join. Can you give me more information regarding the subjects or what can I contribute to the conversation? My email is anhmdq99@gmail.com.

PR #678, opened 2024-01-31
Add linl33 v2

linl33:
#### Check List:

* Execution time: 3.5s, ~100ms improvement compared to previous version

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -53,8 +53,13 @@ jobs:
         id: sdkman
 
       - name: 'Build project'
+        shell: bash
         run: |
           source "$HOME/.sdkman/bin/sdkman-init.sh"
+          if [ -f ${{ format('src/main/java-22/dev/morling/onebrc/CalculateAverage_{0}.java', github.event.pull_request.user.login || '') }} ]; then
+            sdk install java 22.ea.32-open || true
+            sdk use java 22.ea.32-open
+          fi
           ./mvnw --version
           ./mvnw -B clean verify -Pci
 
@@ -63,5 +68,3 @@ jobs:
         run: |
           ./test_ci.sh ${{ github.event.pull_request.user.login }}
         if: github.event_name == 'pull_request'
-
-
--- a/prepare_linl33.sh
+++ b/prepare_linl33.sh
@@ -17,8 +17,7 @@
 
 
 source "$HOME/.sdkman/bin/sdkman-init.sh"
-# TODO: bump to ea 32 when available
-sdk use java 22.ea.31-open 1>&2
+sdk use java 22.ea.32-open 1>&2
 
 CLASS_NAME="CalculateAverage_linl33"
 
--- a/src/main/java-22/dev/morling/onebrc/CalculateAverage_linl33.java
+++ b/src/main/java-22/dev/morling/onebrc/CalculateAverage_linl33.java
@@ -71,7 +71,7 @@ public class CalculateAverage_linl33 {
             final var inputMapped = channel.map(FileChannel.MapMode.READ_ONLY, 0, channel.size(), Arena.global());
 
             final var chunkBounds = calcChunkBounds(inputMapped.address(), inputMapped.byteSize());
-            final var maps = new SparseMap[N_THREADS];
+            final var maps = new HashTable[N_THREADS];
 
             try (final var threadPool = Executors.newFixedThreadPool(N_THREADS, THREAD_BUILDER.factory());
                     final var singleThreadExecutor = Executors.newSingleThreadExecutor(Thread.ofVirtual().factory())) {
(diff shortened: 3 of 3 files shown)


.github/workflows/maven.yml
```diff
@@ -53,8 +53,13 @@ jobs:
         id: sdkman
 
       - name: 'Build project'
+        shell: bash
         run: |
           source "$HOME/.sdkman/bin/sdkman-init.sh"
+          if [ -f ${{ format('src/main/java-22/dev/morling/onebrc/CalculateAverage_{0}.java', github.event.pull_request.user.login || '') }} ]; then
+            sdk install java 22.ea.32-open || true
+            sdk use java 22.ea.32-open
+          fi
```
gunnarmorling, line 62: This should not be needed, the `evaluate.sh` script takes care of installing all required JDKs.

gunnarmorling (commented):
It's faster for the standard key set, but takes unduly lon for 10K keys (see create_measurements_3.sh). Could you take a look, would be sad to see it with DNF on the 10K leaderboard (I am evaluating the top 25 entries against that key seet).

linl33:
The 10k dataset takes ~40s on my laptop (6 cores, no HT) now. Still not fast but I don't think there's anything quick I can do to fix it now.

.github/workflows/maven.yml
linl33: `test_ci.sh` installs the JDKs but it runs after maven so without that line, my class under `java-22/` wouldn't be compiled.

Here's the original run without the change
https://github.com/gunnarmorling/1brc/actions/runs/7724135717/job/21055662835

tivrfoa:
> It's faster for the standard key set, but takes unduly lon for 10K keys (see create_measurements_3.sh). Could you take a look, would be sad to see it with DNF on the 10K leaderboard (I am evaluating the top 25 entries against that key seet).

Hi @gunnarmorling !

One suggestion for when you run the 10k:
- make 5000 cities with < 50 bytes
- **make 5000 cities with >= 50 bytes**
- have at least one city with 100 bytes

Because the 10k is what this challenge is really asking.
But many solutions took advantage of the small dataset, and most cities are less than 16 bytes ...

The top 10 for the 10k might be really different. =)

gunnarmorling:
> One suggestion for when you run the 10k

It's pretty much doing that (see create_measurements_3.sh which creates that file).

> Because the 10k is what this challenge is really asking.
> But many solutions took advantage of the small dataset, and most cities are less than 16 bytes ...

Yes, agreed. I had not anticipated this, and in hindsight I should have not exposed the specific key set. But by the time folks started to optimize for the 413 key set and I realized it, I felt it was too late for changing the evaluation key set, as it would have created lots of frustration. In the end, people did what they were asked to do, so that's totally fair :)

> The top 10 for the 10k might be really different. =)

Yes, which is why there is that separate leaderboard for the 10K keyset which I update occassionally by running the top 25 or so entries from the regular leaderboard against that keyset. Now there could be entries which score lower on the standard leaderboard but would be in the top 25 for 10K, but I felt that's acceptable, given it's just a side eval. Learning a lot myself on how to run these kinds of things :)

gunnarmorling:
Thanks a lot for the update, @linl33. This looks good now. A bit faster on the standard key set and also comes in at ~19 sec for 10K with the correct results. Thx for participating in 1BRC and creating one of the top entries!

```
Benchmark 1: timeout -v 300 ./calculate_average_linl33.sh 2>&1
  Time (mean ± σ):      2.810 s ±  0.056 s    [User: 17.796 s, System: 0.807 s]
  Range (min … max):    2.670 s …  2.865 s    10 runs

Summary
  linl33: trimmed mean 2.8200656749800004, raw times 2.8377631459800003,2.85637806698,2.67035041898,2.8544735929800003,2.8047393789800004,2.86473951498,2.81122164398,2.8182784189800003,2.77623454898,2.80143660298

Leaderboard

|   | 00:02.820 | [link](https://github.com/gunnarmorling/1brc/blob/main/src/main/java/dev/morling/onebrc/CalculateAverage_linl33.java)| 22.ea.32-open | [Li Lin](https://github.com/linl33) | uses Unsafe |
```

linl33:
@gunnarmorling Thank you for organizing this event!

gunnarmorling:
Hey @linl33!

Congrats again on being in the Top 20 of the One Billion Row Challenge!

To celebrate this amazing achievement, I would like to send you a 1BRC t-shirt and coffee mug. To claim your prize, fill out [this form](https://docs.google.com/forms/d/e/1FAIpQLScQLlJiI0KU8rMgoWaNAyKNO_O2Ndmjq9qnv_vmNq6rpL87LA/viewform) by Feb 18. **After** submitting the form, please provide a comment with the random value you've specified in the form, so that I know it is you who submitted it.

All data entered will solely be used in relation to processing this shipment. Shipments can be sent to any country listed [here](https://help.spreadshirt.com/hc/de/articles/207165489-Lieferzeit-Versandkosten?locale=de_DE) or [here](https://help.spreadshirt.com/hc/en-us/articles/207165489-Shipping-Times-and-Costs) (I'll use whichever one is cheaper for me to ship to your location). A big thank you to [Decodable](https://www.decodable.co/) for sponsoring these prizes!

Thanks a lot for participating in 1BRC,

--Gunnar

linl33:
Hi @gunnarmorling,

My 2 random numbers are
```
12533 32231
```

PR #692, opened 2024-01-31
Shipilev: improve comments

shipilev:
This is an easy one, it does not change code, just amends the comments. I have done light testing with it, but there should be no regressions, because nothing executable was changed.


gunnarmorling:
Nice one, thanks a lot! Great spirit!

PR #705, opened 2024-01-31
Use crac to start faster

ianopolousfast:
@gunnarmorling I'm trying to get crac working with zulu jvm, but the maven build won't find jdk.crac package. Do I need to change the pom.xml? I can compile and run it fine in IntelliJ.

(GraalVM native-image doesn't support the vector api, hence this route)

### DIFF
--- a/prepare_ianopolousfast.sh
+++ b/prepare_ianopolousfast.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 #  Copyright 2023 The original authors
 #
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Yes, it will have to be added to the configuration of `maven-compiler-plugin` in _pom.xml_. It would have to be a new profile similar to the one for JDK 22, enabled based on that Zulu JDK (you can activate based on vendor string, I believe). Note that we're after the cut-off time, you'll have two more changes you can make to this PR (see note at the top of the README). If it's not working or valid then, I'll have to close it unfortunately. Updates should be pushed quickly, so I can evaluate all the pending entries. Thx!

ianopolousfast:
I can't get it working in maven after an hour so I think I'm going to give up. Shame, as it would have evened the playing field between AOT and JVM based runs.

ianopolousfast:
Maybe next year... ;-)

gunnarmorling:
Gotcha, thanks for trying. Would also be very interested in seeing results
with CRaC. Although someone said they tried and it didn't yield much if
even any benefit. Thanks a lot for participating in 1BRC!

PR #723, opened 2024-02-26
Naive Ruby & Go implementations

jbarrieault:
Trying a naive implementation in both Ruby and Go.

The implementation:
- creates a map of city -> min/max/count/sum
- reads the measurements file line by line
- sets/updates city key in the map using the line temperature data
- loops over the completed city map key/vals, calculating mean using the data then prints the result

Everything is serial.


jbarrieault:
😳

PR #725, opened 2024-03-03
More accurate file size estimate in create_measurements.py

danielpatrickdotdev:
The current estimate says it could be around twice the actual file size, and I was curious if I could improve on that.

This code should result in a fairly accurate estimate of the actual size (within ~1% from what I've seen). It'll never be perfect because of the random nature of the script, but maybe this is close enough?


gunnarmorling (approved):

gunnarmorling:
LGTM, merged. Thx, @danielpatrickdotdev!

PR #753, opened 2024-08-20
file encoding to utf8

kimmolinna:


JulianStremel:
came here to pull the same :)

